PR #7, opened 2017-08-24
Adding new database schema and patches, and updated relevant points

josenavas:
Instead of creating a different branch in Qiita (which can potentially create merge conflicts down the line) I've decided to keep the DB documentation and generation code independent until we are ready to deploy this in production. At that point we can port all this information to the Qiita repository so everything will be centralized.

In this PR I've added the dbs and html files of the labman DB schema, the file created by dbschema to create this database (db_patch.sql) and a file with the manual editions to the SQL that I had to do so all the schema is sound (db_patch_manual.sql). I've added them in 2 files, so if we need to update db_patch.sql we don't need to remember doing any manual changes, since those are summarized in a different file.

Finally I've also updated the installation instructions, travis and the reset of the DB.


labman/db/testing.py
```diff
@@ -34,3 +36,15 @@ def setUpClass(cls):
             "https://localhost:21174", client_id, client_secret,
             server_cert=labman_settings.qiita_server_cert)
         qclient.post("/apitest/reset/")
+        # The above call only resets the qiita schema. Reset here the labman
```
jdereus: it was my understanding that we would be joining the labman (Bob) schema in to the qiita db schema.  making them one single db (and schema).  is this just for testing currently?

josenavas: I may be misremembering. I can add into the same Qiita schema. I will update the PR accordingly.

PR #9, opened 2017-10-03
Plating gui

josenavas:
This adds the interface to perform the plating. The plate connections with the DB are not in place, but the webserver endpoints are implemented (with hard-coded results).  Once the plate object(s) are in place those placeholder functions can be deleted and replaced with the actual DB calls.

I've added a fair amount of libraries hence the size of the PR!


jdereus:
build fails with error
The command "ipython profile create qiita-general --parallel" failed and exited with 1 during .

jdereus:
build fails

redis.exceptions.ConnectionError: Error 111 connecting to localhost:7777. Connection refused.
The command "qiita-env make --no-load-ontologies" failed and exited with 1 during .

jdereus:
can not find mock.

also The command "nosetests --with-doctest --with-coverage -v --cover-package=labman" exited with 1.

PR #14, opened 2017-12-21
Sample plating

josenavas:
This connects the plating interface with the backend - so that page is fully functional. The rest of the pages are going to be coming as the day moves forward. The amount of changes are pretty high here, so I'm not sure how Github is going to react on the diff...


josenavas:
plating

PR #55, opened 2018-01-17
Normalization

antgonza:


antgonza:
This is ready for review.

labman/gui/templates/normalization.html
```diff
@@ -0,0 +1,183 @@
...
+        'water': $('#water-input').val(),
+        'plate_name': $('#plate-name').val(),
+        'total_vol': $('#total-volume').val(),
+        'ng': $('#ng-input').val(),
+        'min_vol': $('#min-input').val(),
+        'max_vol': $('#max-input').val(),
+        'resolution': $('#resolution-input').val(),
+        'reformat': $('#reformat-input').val()
+      }, function(data) {
+      bootstrapAlert('Information saved', 'success');
```
josenavas: After discussion with the group, we think that after a successful post, we need to "disable" all the inputs from the HTML, and show a button to "Download pick-list". The class NormalizationProcess has a function generate_echo_picklist to generate that file on the fly. That would need to have it's own endpoint, which can be "/process/normalize/([0-9]+)/echo_pick_list$" that returns the file. An example of such handler can be seen here: https://github.com/jdereus/labman/blob/master/labman/gui/handlers/process_handlers/sequencing_process.py#L46

Also, it would be ideal if these tasks can be encapsulated in a function, since we will be able to re-use it later.

antgonza: k, working on this.

antgonza:
This should be ready for review. Note that the interface works as expected, however from my tests there are cases where the id returned is not valid for download because that plate hasn't been fully process or something like that, which is out the scope of this PR but wanted to bring it to @josenavas attention. The error is:
```python
  File "/Users/antoniog/svn_programs/labman/labman/db/process.py", line 1040, in generate_echo_picklist
    composition = well.composition
AttributeError: 'NoneType' object has no attribute 'composition'
```

Issue #54, opened 2018-01-16
Update plate condense to allow for one plate


tanaes:

josenavas:
Fixed in #67

PR #67, opened 2018-01-17
updated plate compression logic

tanaes:
This improves plate compression logic and allows single plates to be compressed to 384-well


tanaes:
Per issue #54

PR #71, opened 2018-01-18
WIP: Library pooling interface

tanaes:
This is pretty rough. I stubbed out some Pooling Handler classes to handle 'trial' pooling and 'commit' pooling for user visualization and approval, respectively. Semar has a functional interface for choosing between library types and creating parameter input dialogues contingent on library type. They need to be finished, connected, and tested.


tanaes:
Most recent commit should have functional handlers, needs:
- [ ] testing of handlers
- [ ] html updated to get and receive correct values from handlers

josenavas:
Commits have been included in #85

Issue #33, opened 2018-01-09
Duplicate Sample ID


ghsmu414:
When entering samples into the sample plate map, an error does not occur when two wells are labeled with the same ID.  Even if the sample is loaded twice, each well needs a unique identifier and a way to relate it back to the metadata.

josenavas:
Fixed in #78

PR #78, opened 2018-01-19
Duplicates and search

josenavas:
This adds the functionality to the plate object to search for plates based on different parameters and the ability to find the duplicated samples in a plate. Note, this is just backend - no interface yet.

### DIFF
--- a/labman/db/plate.py
+++ b/labman/db/plate.py
@@ -107,6 +107,68 @@ class Plate(base.LabmanObject):
     _id_column = "plate_id"
 
     @staticmethod
+    def search(samples=None, plate_notes=None, well_notes=None,
+               query_type='INTERSECT'):
+        """Search plates
+
+        Parameters
+        ----------
+        samples: list of str, optional
+            The samples to find in the plates
+        plate_notes : str, optional
+            The plate notes string to search for. Default: None
+        well_notes : str, optional
+            The well notes string to search for. Default: None
+        query_type : {INTERSECT, UNION}
+            Whether to return the results that fullfill all of the search
+            restrictions or just one of them. Defaul: INTERSECT
+        """
+        if samples is None and plate_notes is None and well_notes is None:
+            raise ValueError(
+                'Please, provide at least on of samples, plate_notes or '
+                'well_notes to perform a plate search.')
+
+        if query_type not in ('INTERSECT', 'UNION'):
+            raise ValueError('query_type should be INTERSECT or UNION. Found:'
+                             ' %s' % query_type)
+
+        with sql_connection.TRN as TRN:
+            sql_queries = []
+            sql_args = []
+            if samples:
+                sql_queries.append(
+                    """SELECT DISTINCT plate_id
+                       FROM qiita.well
+                            JOIN qiita.composition USING (container_id)
+                            JOIN qiita.sample_composition
+                                USING (composition_id)
+                       WHERE sample_id IN %s""")
+                sql_args.append(tuple(samples))
+            if plate_notes:
+                sql_queries.append(
+                    """SELECT plate_id
+                       FROM qiita.plate
+                       WHERE notes @@ to_tsquery(%s)""")
+                sql_args.append(
+                    ' & '.join([w for w in plate_notes.split(' ') if w]))
+            if well_notes:
+                sql_queries.append(
+                    """SELECT plate_id
+                       FROM qiita.well
+                            JOIN qiita.composition USING (container_id)
+                       WHERE notes @@ to_tsquery(%s)""")
+                sql_args.append(
+                    ' & '.join([w for w in well_notes.split(' ') if w]))
+
+            if len(sql_queries) > 1:
+                sql = (" %s " % query_type).join(sql_queries)
+            else:
+                sql = sql_queries[0]
+            TRN.add(sql, sql_args)
+            res = [Plate(pid) for pid in TRN.execute_fetchflatten()]
+        return res
+
+    @staticmethod
     def list_plates(plate_type=None):
         """Generates a list of plates with some information about them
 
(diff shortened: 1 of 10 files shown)


labman/db/plate.py
```diff
@@ -106,6 +108,68 @@ class Plate(base.LabmanObject):
...
+            The plate notes string to search for. Default: None
+        well_notes : str, optional
+            The well notes string to search for. Default: None
+        query_type : {INTERSECT, UNION}
+            Whether to return the results that fullfill all of the search
+            restrictions or just one of them. Defaul: INTERSECT
+        """
+        if samples is None and plate_notes is None and well_notes is None:
+            raise ValueError(
+                'Please, provide at least on of samples, plate_notes or '
```
antgonza: Can you rewrite?

labman/db/plate.py
```diff
@@ -106,6 +108,68 @@ class Plate(base.LabmanObject):
...
+                                USING (composition_id)
+                       WHERE sample_id IN %s""")
+                sql_args.append(tuple(samples))
+            if plate_notes:
+                sql_queries.append(
+                    """SELECT plate_id
+                       FROM qiita.plate
+                       WHERE notes @@ to_tsquery(%s)""")
+                sql_args.append(
+                    ' & '.join([w for w in plate_notes.split(' ') if w]))
```
antgonza: Should this be just split(), this will remove the need of if w.

labman/db/plate.py
```diff
@@ -106,6 +108,68 @@ class Plate(base.LabmanObject):
...
+            if well_notes:
+                sql_queries.append(
+                    """SELECT plate_id
+                       FROM qiita.well
+                            JOIN qiita.composition USING (container_id)
+                       WHERE notes @@ to_tsquery(%s)""")
+                sql_args.append(
+                    ' & '.join([w for w in well_notes.split(' ') if w]))
```
antgonza: Should this be just split(), this will remove the need of if w.

labman/db/plate.py
```diff
@@ -268,6 +336,28 @@ def quantification_process(self):
...
+        with sql_connection.TRN as TRN:
+            sql = """SELECT sample_id, array_agg(well_id ORDER BY well_id)
+                     FROM qiita.well
+                        JOIN qiita.composition USING (container_id)
+                        JOIN qiita.sample_composition USING (composition_id)
+                     WHERE sample_id IS NOT NULL AND plate_id = %s
+                     GROUP BY sample_id
+                     HAVING array_length(array_agg(well_id), 1) > 1"""
+            TRN.add(sql, [self.id])
+            res = {sample_id: [container_module.Well(w) for w in wells]
```
antgonza: Why do this versus getting everything in a single command. In other words, I'm not sure in the benefit of doing this vs a larger SQL with array_agg

josenavas:
This now includes the modifications needed in the plating interface (I didn't realize that I was working off the same branch). Amanda is working on the searching interface.

Fixes: #33

labman/db/plate.py
```diff
@@ -300,3 +390,59 @@ def get_well(self, row, column):
...
+        """
+        with sql_connection.TRN as TRN:
+            sql = """SELECT well_id
+                     FROM qiita.well
+                        JOIN qiita.composition USING (container_id)
+                        JOIN qiita.sample_composition USING (composition_id)
+                     WHERE plate_id = %s AND sample_id = %s
+                     ORDER BY well_id"""
+            TRN.add(sql, [self.id, sample_id])
+            res = [container_module.Well(well)
```
antgonza: Same here, I think it not only will be faster but also cleaner.

labman/db/plate.py
```diff
@@ -300,3 +390,59 @@ def get_well(self, row, column):
...
+                        JOIN qiita.well USING (container_id)
+                     WHERE plate_id <> %s AND sample_id IN (
+                        SELECT sample_id
+                        FROM qiita.sample_composition
+                            JOIN qiita.composition USING (composition_id)
+                            JOIN qiita.well USING (container_id)
+                            WHERE plate_id = %s)
+                     GROUP BY plate_id"""
+            TRN.add(sql, [self.id, self.id])
+            prev_plated = TRN.execute_fetchindex()
```
antgonza: Same here. A little more complex SQL will avoid multiple queries.

labman/db/plate.py
josenavas: Changed

labman/db/plate.py
josenavas: I actually need the if to remove empty strings:
```ipython

In [2]: s = 'aaa    bbbb;'

In [3]: s.split(' ')
Out[3]: ['aaa', '', '', '', 'bbbb;']
```

labman/db/plate.py
josenavas: Same as above

labman/db/plate.py
josenavas: The SQL is already doing the array_agg, this is actually construction the object, rather than just returning the ID.

labman/db/plate.py
josenavas: Same as above - it is just constructing the object.

labman/db/plate.py
josenavas: In this case that is true, it is performing multiple queries.

However, I've not been able to come up with a SQL query that achieves what I want here - I'm open for ideas.

labman/db/plate.py
antgonza: K, thanks.

labman/db/plate.py
antgonza: K, thanks.

labman/db/plate.py
antgonza: K, let's leave it like this and we can improve later if it becomes a bottleneck.

josenavas:
@antgonza doesn't have merge privileges - so I'm going to merge to not block dev.

PR #82, opened 2018-01-20
Fix sequencing process

josenavas:
We can sequence a pool per lane - fix the sequencing process to store that information.


josenavas:
Merging so I can keep moving forward

Issue #76, opened 2018-01-19
Provide an interface that, given a sample id, find all plates that the sample is present


josenavas:
Similarly, given a comment (or key words) find the plates where the comment is found.

josenavas:
Providing a little bit more information on how this page should look. The interface should present to the user with 3 input elements (input as general concept not actual <input> tag :) ).

The first input element should allow the user to provide a list of samples. Ideally it will work like the sample plating page, in which the user first chooses the study so the system can provide a list of samples for such study.

The second input element should be a free text that will search over all the PLATE comments.

The third input element should be a free text that will search over all the WELL comments.

Finally, it should present a dropdown menu with 2 options: INTERSECT and UNION (with default to INTERSECT).
Probably adding a comment would be useful to explain each of them on the interface. INTERSECT will retrieve all the plates in which the samples AND the comments are found. UNION will retireve the plates in which the samples OR the comments are found.

The result would be a table like the one in http://localhost:8080/plates, so we could potentially use the same table structure but populate the results differently.

josenavas:
Fixed in #92

PR #83, opened 2018-01-20
Interface to find sample plates on which specified sample(s) are present

AmandaBirmingham:
Implements functionality for #76

However, here's what's missing:
1) verification that the call in plate.py to Plate.search works, as that function has been written but not yet merged into main
2) implementation of requirement that user should not be able to enter samples to search without first adding at least one study to search (code currently lets users search ALL studies)

Additionally, I suggest that there should be another validation added: currently the javascript ensures that at least one sample input has been added--but doesn't verify that the sample input has a non-empty-string value.  D'oh :)

### DIFF
--- a/labman/gui/handlers/plate.py
+++ b/labman/gui/handlers/plate.py
@@ -37,6 +37,25 @@ def _get_plate(plate_id):
     return plate
 
 
+class PlateSearchHandler(BaseHandler):
+    @authenticated
+    def get(self):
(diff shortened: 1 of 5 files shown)


AmandaBirmingham:
Since this pull request hasn't been processed yet, I'm just going to close it and make a new one from my updated branch now that I've been added as a collaborator.

PR #84, opened 2018-01-20
Finishing up the Sequencing interface

josenavas:
This includes the interface originally developed by @qiyun, but I fixed the latest bits.


qiyunzhu:
@josenavas good job, thanks!

PR #85, opened 2018-01-22
Moving forward on the pooling page

josenavas:
This includes the commits from #71


labman/gui/handlers/process_handlers/quantification_process.py
```diff
@@ -27,46 +27,36 @@ def get(self):
...
+        # plate-file-<PLATE_ID> so use the keys to know the plates
+        # that we need to quantify
+        plates = []
+        for key in self.request.files:
+            plate_id = key.rsplit('-', 1)[1]
+            # The 0 is because for each key we have a single file
+            file_content = self.request.files[key][0]['body'].decode('utf-8')
+            plate = Plate(plate_id)
+            pc = plate.plate_configuration
+            concentrations = QuantificationProcess.parse(
```
wasade: what happens if one of the files is malformed, should this be wrapped in a try/except? would a rollback be necessary?

labman/db/process.py
```diff
@@ -1567,6 +1554,102 @@ def concentrations(self):
...
+                row = well.row - 1
+                col = well.column - 1
+                sample_concs[row][col] = r_conc
+                sc = comp.gdna_composition.sample_composition
+                is_blank[row][col] = sc.sample_composition_type == 'blank'
+
+            res = QuantificationProcess._compute_amplicon_pico_concentration(
+                sample_concs, dna_amount)
+            res[is_blank] = blanks
+            res[sample_concs < min_val] = min_val
```
tanaes: It doesn't seem appropriate to me to store incorrect values under a 'quantification process.' Instead, the best estimate of values should be stored, and pooling handled depending on those values.

labman/db/process.py
```diff
@@ -1567,6 +1554,102 @@ def concentrations(self):
...
+                        SET computed_concentration = %s
+                        WHERE upstream_process_id = %s AND
+                              quantitated_composition_id = %s"""
+
+            with sql_connection.TRN as TRN:
+                TRN.add(sql, sql_args, many=True)
+                TRN.execute()
+
+    @staticmethod
+    def _compute_amplicon_pico_concentration(sample_concs, dna_amount=240):
```
tanaes: This isn't computing concentrations, it's computing pooling volumes.

labman/gui/handlers/process_handlers/pooling_process.py
```diff
@@ -7,12 +7,155 @@
...
+        plate_info = json_decode(self.get_argument('plate-info'))
+
+        plate_id = plate_info['plate-id']
+        func_name = plate_info['pool-func']
+        func_info = POOL_FUNCS[func_name]
+        function = func_info['function']
+        params = {arg: plate_info['%s%s' % (pfx, plate_id)]
+                  for arg, pfx in func_info['parameters']}
+
+        plate = Plate(plate_id)
```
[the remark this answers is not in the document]
tanaes: Or is this getting individual post requests from the webpage and returning one plate at a time?

labman/gui/handlers/process_handlers/test/test_pooling_process.py
```diff
@@ -0,0 +1,37 @@
...
+# ----------------------------------------------------------------------------
+
+from unittest import main
+
+from tornado.escape import json_decode, json_encode
+
+from labman.gui.testing import TestHandlerBase
+
+
+class TestPoolingProcessHandlers(TestHandlerBase):
```
wasade: would it be possible to add in failure mode tests?

labman/gui/templates/library_pooling.html
```diff
@@ -0,0 +1,276 @@
+{% extends sitebase.html %}
+
+{% block head %}
+<script type='text/javascript'>
+
+  var poolParams = {
```
wasade: are these defaults assured to be in sync with the python library code? is there a mechanism in place to ensure they are consistent? if not, there will be unexpected behavior in the future if and when this block (or the python code) under goes maintenance.

labman/gui/handlers/process_handlers/pooling_process.py
```diff
@@ -7,12 +7,155 @@
...
+        quant_process = plate.quantification_process
+
+        output = {}
+        if func_name == 'Amplicon':
```
tanaes: I think it makes more sense to make 'amplicon' a pooling function. We should also expose 'equal volume' pooling to amplicon plates. Really all that needs to happen is *concentration* calculations performed separately depending on library type (this gives us the flexibility down the road of exposing alternative concentration calculation methods, like qPCR), and the pooling logic be separated from the conc calc into the pooling function.

labman/gui/handlers/process_handlers/pooling_process.py
```diff
@@ -7,12 +7,155 @@
...
+            # Amplicon
+            # Compute the normalized concentrations
+            quant_process.compute_concentrations(**params)
+            # Compute the pooling values
+            output['pool_vals'] = make_2D_array(plate, quant_process)
```
tanaes: It doesn't appear that there is any pooling function being applied here.

labman/gui/templates/library_pooling.html
```diff
@@ -0,0 +1,276 @@
...
+                 {'prefix': 'min-val-', 'value': '1', 'desc': 'Minimum concentration value (ng/&mu;l):', 'min': '0.001', 'step': '0.001'},
+                 {'prefix': 'max-val-', 'value': '15', 'desc': 'Maximum concentration value (ng/&mu;l):', 'min': '0.001', 'step': '0.001'},
+                 {'prefix': 'blank-val-', 'value': '2', 'desc': 'Blanks value:', 'min': '0.001', 'step': '0.001'}]};
+
+  function getPlateInformation(elem) {
+    var plateId = $(elem).attr('pm-data-plate-id');
+    var plateType = $('#plate-type-select').val();
+    var parameterName, poolFunc;
+
+    if (plateType === '16S library prep') {
```
wasade: how difficult will it be to add in 18S and ITS preparations into labman? is it obvious what all of the changes necessary are?

labman/gui/templates/parse_quantification.html
```diff
@@ -7,11 +7,88 @@
...
+  };
+
+  function addPlate(plateId) {
+    $.get('/plate/' + plateId + '/', function (data) {
+      var $divElem = $("<div>");
+      $divElem.addClass('list-group-item');
+      $divElem.attr('id', 'plate-' + plateId);
+      $divElem.attr('pm-data-plate-id', plateId);
+      $divElem.append('<label><h4>' + data.plate_name + '</h4></label>');
+      $divElem.append(' (' + data.plate_configuration[1] + ')');
```
wasade: why `[1]`?

labman/gui/templates/parse_quantification.html
```diff
@@ -7,11 +7,88 @@
```
wasade: it seems like there is a fair bit of overlap with `library_pooling.html`, should the scripting logic be placed into a common library?

labman/gui/templates/quantification.html
```diff
@@ -28,40 +29,46 @@
...
+      for (var i = 0; i < concentrations[0].length; i++) {
+        sgCols.push({id: i, name: i+1, field: i});
+      }
+      var data = [];
+      var rowId = 'A';
+      for (var i = 0; i < concentrations.length; i++) {
+        var d = (data[i] = {});
+        d["header"] = rowId;
+        for (var j = 0; j < concentrations[i].length; j++) {
+          d[j] = concentrations[i][j];
```
wasade: the python code uses 1-based indexing for row and column positions, but it looks as though this is using 0-based. is the client-side javascript code really using a different indexing scheme for well positions than server side? if yes, then that seems like a critical issue.

labman/gui/handlers/process_handlers/pooling_process.py
```diff
@@ -7,12 +7,155 @@
...
+        else:
+            # Shotgun
+            # Compute the normalized concentrations
+            size = params.pop('size')
+            quant_process.compute_concentrations(size=size)
+            # Compute the pooling values
+            sample_concs = make_2D_array(plate, quant_process)
+            output['pool_vals'] = function(sample_concs, **params)
+
+        self.write(output)
```
tanaes: Why are we passing back this numpy array in the `write`? Shouldn't we be computing some sort of summary representation for the user (e.g. running the matplotlib image generation)? Or are we intending to have client-side rendering of this information?

labman/db/tests/test_process.py
```diff
@@ -864,27 +869,27 @@ def test_compute_shotgun_pooling_values_eqvol(self):
             [[98.14626462, 487.8121413, 484.3480866, 2.183406934],
              [498.3536649, 429.0839787, 402.4270321, 140.1601735],
              [21.20533391, 582.9456031, 732.2655041, 7.545145988]])
-        obs_sample_vols = PoolingProcess._compute_shotgun_pooling_values_eqvol(
+        obs_sample_vols = PoolingProcess.compute_shotgun_pooling_values_eqvol(
             qpcr_conc, total_vol=60.0)
         exp_sample_vols = np.zeros([3, 4]) + 60.0/12*1000
```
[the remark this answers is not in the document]
tanaes: should be 5000 (nL)

labman/db/tests/test_process.py
```diff
@@ -732,30 +732,35 @@ def test_attributes(self):
...
+        self.assertEqual(obs_c[0][0], LibraryPrep16SComposition(1))
         npt.assert_almost_equal(obs_c[0][1], concentrations[0][0])
         self.assertIsNone(obs_c[0][2])
-        self.assertEqual(obs_c[12][0], GDNAComposition(61))
+        self.assertEqual(obs_c[12][0], LibraryPrep16SComposition(13))
         npt.assert_almost_equal(obs_c[12][1], concentrations[1][0])
         self.assertIsNone(obs_c[12][2])
+        obs.compute_concentrations()
+        obs_c = obs.concentrations
+        self.assertIsNotNone(obs_c[0][2])
```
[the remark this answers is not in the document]
josenavas: Good call - added.

labman/db/tests/test_process.py
```diff
@@ -732,30 +732,35 @@ def test_attributes(self):
...
         self.assertEqual(obs.date, date.today())
         self.assertEqual(obs.personnel, user)
         obs_c = obs.concentrations
         self.assertEqual(len(obs_c), 384)
         self.assertEqual(obs_c[0][0], LibraryPrepShotgunComposition(1))
         npt.assert_almost_equal(obs_c[0][1], concentrations[0][0])
+        self.assertIsNone(obs_c[0][2])
+        obs.compute_concentrations()
+        obs_c = obs.concentrations
         self.assertIsNotNone(obs_c[0][2])
```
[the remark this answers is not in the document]
josenavas: Added

labman/db/tests/test_process.py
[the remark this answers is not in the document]
josenavas: Added the 5000 instead of the computation for clarity.

labman/gui/handlers/process_handlers/pooling_process.py
```diff
@@ -7,12 +7,155 @@
...
 from labman.gui.handlers.base import BaseHandler
+from labman.db.process import PoolingProcess, QuantificationProcess
+from labman.db.plate import Plate
+from labman.db.equipment import Equipment
+from labman.db.composition import PoolComposition
 
 
-class PoolProcessHandler(BaseHandler):
+# quick function to create 2D representation of well-associated numbers
+def make_2D_array(wells, vals):
```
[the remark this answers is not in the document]
josenavas: This function and the next ones are 2 functions that I didn't modified from Jon's original PR, as I was still finishing up with that part of the interface. Now that I've completed the PR this has been fixed.

labman/gui/handlers/process_handlers/pooling_process.py
```diff
@@ -7,12 +7,155 @@
...
+    val_array = np.zeros_like(x.layout, dtype=float) + np.nan
```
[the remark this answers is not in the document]
josenavas: That's what I was wondering...

labman/gui/handlers/process_handlers/pooling_process.py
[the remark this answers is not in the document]
josenavas: This is getting individual requests and returning one plate at a time. It will be more efficient and conceptually cleaner.

labman/gui/handlers/process_handlers/pooling_process.py
josenavas: According to Greg, the normalized concentrations is what they as pooling values.

labman/gui/handlers/process_handlers/pooling_process.py
josenavas: Current request from the 16S involves visualizing the values - I think once your code to plot these values is ported over, we can change this (and adapt GUI accordingly). However, that is deferred as it is not critical functionality and current values is what Greg asked for.

labman/gui/handlers/process_handlers/quantification_process.py
josenavas: There is no modification of the DB in this section of the code (i.e. no rollback necessary).
If the file is malformed an error is raised, then tornado handles it and it is shown to the user in the bootrstapAlert

labman/gui/handlers/process_handlers/test/test_pooling_process.py
josenavas: Added where possible

labman/gui/templates/library_pooling.html
josenavas: 18S and ITS preparations are already supported - The interface exposes everything as 'Amplicon' instead of 16S, but the entire DB and backend uses '16S' instead of amplicon. Fixing this should be simple, but is tedious since that nomenclature is used everywhere in the system. Since that is not affecting critical functionality, that name change has been deferred for the future. Unfortunately the name error was caught when the project was already really advanced.

labman/gui/templates/parse_quantification.html
josenavas: A lot of the pages use this structure, but they're all slightly different. Although it may be possible to locate the scripting logic in a common library, the modifications that need to be done and the generalization of the code is complex enough that will delay the development too much.

labman/gui/templates/parse_quantification.html
josenavas: Added a comment

labman/gui/templates/quantification.html
josenavas: Yes, they're and this is because of the use of a 3rd party library (slickgrid).

josenavas:
This is no longer a WIP PR.

josenavas:
BTW - by mistake I've been developing this on my master branch (I merged Jon changes on my master branch and it was easier to keep moving on that rather than fixing the branch issues). So any further development that I perform will be done on top of this changes.

labman/gui/templates/library_pooling.html
wasade: Okay, it might be useful to issue this refactor earlier rather than later in the project as it will only get harder do

labman/db/process.py
```diff
@@ -1567,6 +1554,102 @@ def concentrations(self):
...
+                      composition_module.LibraryPrep16SComposition):
+            # Amplicon
+            sample_concs = np.zeros_like(layout, dtype=float)
+            is_blank = np.zeros_like(layout, dtype=bool)
+            for comp, r_conc, _ in concentrations:
+                well = comp.container
+                row = well.row - 1
+                col = well.column - 1
+                sample_concs[row][col] = r_conc
+                sc = comp.gdna_composition.sample_composition
```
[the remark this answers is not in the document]
wasade: oh i see now, okay

labman/gui/templates/library_pooling.html
josenavas: Agree - added an issue that I've assigned to myself: https://github.com/jdereus/labman/issues/91

Happy to address once the minimum functionality required for publication is in place.

labman/db/process.py
```diff
@@ -1567,6 +1557,95 @@ def concentrations(self):
...
+                is_blank[row][col] = sc.sample_composition_type == 'blank'
+
+            res = QuantificationProcess._compute_amplicon_pool_values(
+                sample_concs, dna_amount)
+            res[sample_concs < min_val] = min_val
+            res[sample_concs > max_val] = 0
```
wasade: It sounds like the use of zero is misleading, as well as being a magical number. Would it make sense for the surrounding code to interpret a non-value like `inf` or a negative value (i.e., something that is clearly impossible)?

labman/gui/templates/library_pooling.html
```diff
@@ -0,0 +1,358 @@
...
+
+<script src="/static/vendor/js/jquery.event.drag-2.3.0.js" type="text/javascript"></script>
+<script src="/static/vendor/js/slick.core.js" type="text/javascript"></script>
+<script src="/static/vendor/js/slick.grid.js" type="text/javascript"></script>
+<script src="/static/vendor/js/slick.editors.js" type="text/javascript"></script>
+
+<script type='text/javascript'>
+
+  var poolParams = {
+    'min': [{'prefix': 'floor-vol-', 'value': '100', 'desc': 'volume for low conc samples (nL):', 'min': '1', 'step': '1'},
```
wasade: same comment as before regarding maintaining sync with these parameters with what's in the server side code

josenavas:
@wasade wanna merge?

labman/db/process.py
josenavas: No, the reasoning behind using 0 is that the volume to pool is 0, i.e. do not pool that sample.

Does that make sense?

I will add a comment so the number is not magical anymore.

labman/gui/templates/library_pooling.html
josenavas: I am moving this code to the python handler and adding a unittest that will check the consistency between these two. That is the best that I can do - would that work?

josenavas:
Thanks @wasade I just pushed the latest changes with your comments

josenavas:
@wasade are there any outstanding comments here? If not can you approve the PR for merging? The merge conflicts where on sitebase.html - but I will wait for the tests anyways

josenavas:
Thanks @wasade !!!!!!

PR #87, opened 2018-01-22
popup showing plate controls description

FranckLejzerowicz:
- html change consist in adding a table with the currently 3 controls: (col1) control name (col2) control descrition
- javascript change consist in addind a non-jquery/pure-javascript on click function to show the popup
- labman.ccs was amended for the popup => please change class names at your discretion as I don't know whether other popups will be used.


FranckLejzerowicz:
done

PR #90, opened 2018-01-22
WIP: generate_prep_information

antgonza:
This is the initial code, I would like a review so we can agree on the idea. Note that we need ad hoc SQL cause doing the same thing via objects takes ~420s while 21.070s (including db reset).


labman/db/process.py
```diff
@@ -2380,3 +2381,78 @@ def generate_sample_sheet(self):
...
+                    if study not in data:
+                        data[study] = {}
+                    data[study][sid] = result
+
+        # converting from dict to pandas and then to tsv
+        for study, vals in data.items():
+            df = pd.DataFrame.from_dict(vals, orient='index')
+            cols = sorted(list(df.columns))
+            sio = StringIO()
+            df[cols].to_csv(sio, sep='\t', index_label='sample_name')
```
wasade: I thought the sample sheets were comma separated with a bunch of header information about the sample listings?

antgonza: right but this is actually going to qiita so tsv

josenavas: Note that this is the PrepInformation generation - not the sample sheet :-)

PR #92, opened 2018-01-22
Plate search

AmandaBirmingham:
I believe this completes the work for issue #76, with the single exception that I can't run the unit tests on my system so I can't verify that my gui/test/test_plate.test_post_plate_search_handler test is passing (although I have of course done manual testing :) This test code is currently commented out and needs to be commented in and verified by someone who can run the unit tests ...

### DIFF
--- a/labman/gui/test/test_plate.py
+++ b/labman/gui/test/test_plate.py
@@ -6,7 +6,7 @@
 # The full license is in the file LICENSE, distributed with this software.
 # ----------------------------------------------------------------------------
 
-# from json import dumps
+from json import dumps
 from unittest import main
 
 from tornado.escape import json_decode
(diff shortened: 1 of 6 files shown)


labman/gui/test/test_plate.py
```diff
@@ -6,6 +6,7 @@
 # The full license is in the file LICENSE, distributed with this software.
 # ----------------------------------------------------------------------------
 
+# from json import dumps
```
josenavas: Can you uncomment these tests? They will get executed in travis, so we can see there if they pass or not - and if not, we should be able to see the error.

josenavas:
@AmandaBirmingham it looks like the tests didn't pass - since you're having trouble running the tests locally in your machine, I'm going to solve the tests failures and commit against this branch, so we can move this functionality forward and have it merged.

AmandaBirmingham:
Yeah, I've been looking into the failure.  I think at root it is because the test db has no plates or wells with comments; I tried following the approach that you used in db/tests/test_plate test_search to add some of those comments to plates and wells on the fly so they would be there to test them, but it looks like my simplistic attempt didn't work.  I guess it's up to you whether to extend the test code so it actually does modify the test db so some of those keywords are there to search for, or to prune it back so it only tests that keyword searches (correctly, given the db) return empty plate lists.

PR #93, opened 2018-01-23
Modified handler: no param required + plate search interface

FranckLejzerowicz:
- normalization_process.py: use get_arguments to return a list
- normalization.html:
* added "+Add plate" button and its modal (copy from extraction.html) 
* toggleAddPlate function to disable "+Add plate" button after modal choice (hence, unnecessary addBtnPlate disabling property has been removed)
* added remove plate button and re- toggleAddPlate to re-enable "+Add plate" button

Note1: the normalization page cannot be accessed yet from the navigation menu
Note2: clicking Normalize / normalize_send() throw an error (under development I assume)

### DIFF
--- a/labman/gui/handlers/process_handlers/test/test_normalization.py
+++ b/labman/gui/handlers/process_handlers/test/test_normalization.py
@@ -9,7 +9,7 @@
 from unittest import main
 from tornado.escape import json_decode
 
-from labman.gui.testing import TestHan$lerBase
+from labman.gui.testing import TestHandlerBase
 
 
 class TestNormalizationHandlers(TestHandlerBase):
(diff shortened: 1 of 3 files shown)


josenavas:
@FranckLejzerowicz tests failed here. Can you change the test so it passes? It will be just a matter of changing this 400 [here](https://github.com/jdereus/labman/blob/master/labman/gui/handlers/process_handlers/test/test_normalization.py#L26) for a 200, since now an error is not expected on that page.

Thanks!

labman/gui/handlers/process_handlers/test/test_normalization.py
```diff
@@ -23,7 +23,8 @@ def test_normalization_handler(self):
         self.assertNotEqual(response.body, '')
 
         response = self.get('/process/normalize')
-        self.assertEqual(response.code, 400)
+        print(response)
```
josenavas: Can you remove this print?

PR #95, opened 2018-01-23
Study summary page

qiyunzhu:
Hello @josenavas I need your help in resolving this. I was trying to create a "study.html" page to display study information, but I couldn't get the "view study" button to navigate to this page, after many unsuccessful trials. Perhaps it was some small issue I missed. Here I am posting a blank page with (what I think) necessary linkings. Can you please take a look and let me know what is missing? Thanks!

### DIFF
--- a/labman/gui/handlers/study.py
+++ b/labman/gui/handlers/study.py
@@ -33,16 +33,12 @@ class StudyListHandler(BaseHandler):
 
 class StudyHandler(BaseHandler):
     @authenticated
-    def get(self):
+    def get(self, study_id):
         try:
-            study_id = self.get_argument('study_id', None)
             study = Study(int(study_id))
-            self.render('study.html', study_id=study.id,
-                        study_title=study.title,
-                        total_samples=study.num_samples)
-            # self.write({'study_id': study.id,
-            #             'study_title': study.title,
-            #             'total_samples': study.num_samples})
+            self.write({'study_id': study.id,
+                        'study_title': study.title,
+                        'total_samples': study.num_samples})
         except LabmanUnknownIdError:
             self.set_status(404)
         self.finish()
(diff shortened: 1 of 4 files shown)


josenavas:
Hi @qiyunzhu !

You need to create an endpoint and add it to the webserver. [Here](https://github.com/jdereus/labman/blob/master/labman/gui/webserver.py#L59) is where you can add the endpoint.

Hope this helps!

qiyunzhu:
Hi @josenavas Thanks for instant reply! I tried to use the existing `(r"/study/(.*)/", StudyHandler),` endpoint in `webserver.py`. Does that one not work?

josenavas:
Hi @qiyunzhu 

That handler is used to achieve some functionality in the sample plating page - please do not alter it's response as it will break functionality somewhere else.

You'll need to create a new endpoint, I recommend `"/study/([0-9]+)/sample_summary"`.

Then, modify [this](https://github.com/jdereus/labman/pull/95/files#diff-5b18a77bbbc4a28437b757ba0d91f63bR28) to be

```JavaScript
window.location.href = '/study/' + $(this).attr('data-lb-study-id' + '/sample_summary');
```

Then, in the new handler, define the `get` function as:
```python
@authenticated
def get(self, study_id):
    ....
```
Tornadao will make sure that the number in the regex `"/study/([0-9]+)/sample_summary"` will be passed as a parameter to your get function.

Hope this helps!

qiyunzhu:
@josenavas Cool! Let me try your way. Thanks for the help!

josenavas:
No problem! Thank you!

qiyunzhu:
@josenavas I got the study pages working following your guideline. Now there is a minimum study summary page displaying only sequence title and total number of samples. Once the backend db function is done, more display items can be added.

josenavas:
Thanks @qiyunzhu 
There is a merge conflict - can you address?

Also - ETA for the backend function retrieving the rest of the information?

qiyunzhu:
@josenavas thanks for reminding. Let me resolve the conflict. Pinging @mortonjt who is working on the SQL command.

Issue #101, opened 2018-01-23
Modify processes pages so that they accept a process_id as parameter


josenavas:
If the process id is present, load the pages with the process information pre-loaded, all the inputs disabled and, if there is any file to download, enabling the buttons to download those files (e.g. pooling file in `http://localhost:8080/process/poollibraries`

The pages that need to be modified:
- [x] `http://localhost:8080/process/gdna_extraction` @josenavas #106 
- [x] `http://localhost:8080/process/gdna_compression` @wasade #119 
- [x] `http://localhost:8080/process/normalize` #148
- [x] `http://localhost:8080/process/library_prep_16S` @josenavas #111
- [x] `http://localhost:8080/process/library_prep_shotgun` #148 
- [x] `http://localhost:8080/process/poollibraries` #148
- [x] `http://localhost:8080/process/poolpools` #148

wasade:
I dont have edit permissions, but am starting on the `gdna_compression` page

PR #103, opened 2018-01-23
Fixes #100

### LINKED ISSUE #100: Update plate listing page to provide all types of plates
And add a button per plate that allows to see the plate latest executed process.

josenavas:
the actual redirects from the plate view icon redirect correctly but don't show the information. They will be showing the information once https://github.com/jdereus/labman/issues/101 is fixed.

### DIFF
--- a/labman/gui/templates/plate_list.html
+++ b/labman/gui/templates/plate_list.html
@@ -33,7 +33,7 @@
 
   $(document).ready(function(){
     var table = $('#plateListTable').DataTable(
-      {'columnDefs': [{'targets': 0, 'orderable': false, 'width': '30px'}],
+      {'columnDefs': [{'targets': 0, 'orderable': false, 'width': '50px'}],
        'order': [[1, "desc"]],
        'language': {'zeroRecords': 'No plates found - choose a plate type'}});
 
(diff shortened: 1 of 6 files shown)


labman/gui/templates/plate_list.html
```diff
@@ -91,14 +96,16 @@
...
+    <option value='16S library prep'>Amplicon library plates</option>
+    <option value='shotgun library prep'>Shotgun library plates</option>
   </select>
 </div>
 
 <table id="plateListTable" class="display" cellspacing="0" width="100%">
   <thead>
     <tr>
-      <th></th>
+      <th>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;</th>
```
antgonza: Not blocking but will it be better to set the row width vs adding a bunch of spaces?

josenavas: Good point - modified it on the DataTable object declaration (which already had a value....)

josenavas:
@jdereus able to merge this?

PR #105, opened 2018-01-23
Multiplate - WIP 

semarpetrus:


semarpetrus:
I moved the code to plate.js and that seems to work much better than extending plateViewer. I kept everything as id's but all each plate is in its own div so they are accessible via the id of the div. I have not looked at the python yet, but I will be resuming the work around 2PM.

One main thing that I could not find a clean way to do is generating the list of options for plate configuration. It is done through tornado and I can't copy that in JQuery. My current method to deal with that is to create a hidden version using html and clone that into the plate div and then remove it.

josenavas:
Thanks @semarpetrus 

To solve the problem with the options, the best will be to add a parameter to the JS function that contain those values. Then, on the HTML you can use the {% raw .... %} command to pass those into the call directly.

wasade:
@semarpetrus, is there a plan to continue on this PR or can it be closed?

antgonza:
Closing but feel free to reopen if you want to continue working on this.

PR #110, opened 2018-01-23
BUG: Templates and static were not being included

ElDeveloper:
Non-editable installations would be missing a few directories and files.


wasade:
:+1:

PR #118, opened 2018-01-23
ENH: Adding in queries wrt to number of amplified, quantified, plated…

mortonjt:
…, pooled and sequenced 16S samples

Wasn't able to get the testing database configured properly though ...

### DIFF
--- a/labman/db/study.py
+++ b/labman/db/study.py
@@ -108,3 +108,109 @@ class Study(base.LabmanObject):
                      WHERE study_id = %s"""
             TRN.add(sql, [self.id])
             return TRN.execute_fetchlast()
+
+    @property
+    def num_16S_amplified(self):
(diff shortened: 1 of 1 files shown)


labman/db/study.py
```diff
@@ -108,3 +108,108 @@ def num_samples(self):
                      WHERE study_id = %s"""
             TRN.add(sql, [self.id])
             return TRN.execute_fetchlast()
+
+    @property
+    def num_16S_amplified(self):
```
wasade: The codebase is not actually restricted to 16S amplicons despite the nomenclature, can this be fixed so it's `number_amplicons_amplified`? Same with other instances of the use of 16S

wasade:
...note that the style of the SQL written differs some but that's minor

josenavas:
Closing in favor of #140

PR #119, opened 2018-01-24
WIP: Gdna compression prepopulate

wasade:
Solves prepopulation for this page in #101 

I haven't been able to disable the "remove" buttons dynamically which is frustrating. I'm not sure what other things should be disabled when prepopulating so for right now I'm just disabling everything

### DIFF
--- a/labman/gui/handlers/process_handlers/gdna_compression_process.py
+++ b/labman/gui/handlers/process_handlers/gdna_compression_process.py
@@ -12,13 +12,33 @@ from tornado.escape import json_decode
 from labman.gui.handlers.base import BaseHandler
 from labman.db.process import GDNAPlateCompressionProcess
 from labman.db.plate import Plate
+from labman.db.exceptions import LabmanUnknownIdError
 
 
 class GDNAPlateCompressionProcessHandler(BaseHandler):
     @authenticated
     def get(self):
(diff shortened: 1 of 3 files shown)


josenavas:
Thanks @wasade I'm going to pull down this into my local branch and I will investigate.

josenavas:
I can't reproduce your issue, I think it must be some kind of race condition, which I can imagine where it is originating from. I, however, found a different issue that affects the structure on how we are storing the data of the compressed plates. I'm communicating with @AmandaBirmingham to see which is the best solution moving forward.

wasade:
Okay thanks! I hope the issue isn't too deep

On Jan 24, 2018 2:54 PM, "Jose Navas" <notifications@github.com> wrote:

I can't reproduce your issue, I think it must be some kind of race
condition, which I can imagine where it is originating from. I, however,
found a different issue that affects the structure on how we are storing
the data of the compressed plates. I'm communicating with @AmandaBirmingham
<https://github.com/amandabirmingham> to see which is the best solution
moving forward.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/jdereus/labman/pull/119#issuecomment-360301689>, or mute
the thread

josenavas:
Closing in favor of #139

PR #123, opened 2018-01-24
added more rows to study summary page

qiyunzhu:
Note that the webpage won't work properly at this moment, as the presumed backend functions are not present yet.


josenavas:
Closing in favor of #140

PR #124, opened 2018-01-24
Fixing DB schema error

josenavas:
@AmandaBirmingham brought to my attention that the gDNA extraction process and amplicon library processes where used incorrectly. I introduced a set of tables in January 17 that broke the actual assumption behind those processes, so here I fix the DB schema, fix the objects and the interface to behave as expected.

In short: each process instance applies to a single plate (rather than a set of plates as I had it before).

@AmandaBirmingham this is the DB structure define here, does it correctly match your expectations?

screen shot 2018-01-24 at 2 12 57 pm

### DIFF
--- a/labman/gui/templates/library_prep_16S.html
+++ b/labman/gui/templates/library_prep_16S.html
@@ -9,27 +9,35 @@
 <script type='text/javascript'>
   function prepopulateGUI() {
     var processId = {% raw process_id if process_id else 'null' %};
-    var platesInfo = {% raw plates_info %};
+    var gdnaPlate = '{{gdna_plate}}';
+    var epmotion = '{{epmotion}}';
+    var epmotionTM300 = '{{epmotion_tm300}}';
+    var epmotionTM50 = '{{epmotion_tm50}}';
+    var masterMix = '{{master_mix}}';
+    var waterLot = '{{water_lot}}';
+    var primerPlate = '{{primer_plate}}';
+    var prepDate = '{{preparationDate}}';
+    var volume = '{{volume}}';
+
     if (processId !== null) {
       // We need to prepopulate the interface
       // Start by adding all plates:
-      var requests = [];
-      $.each(platesInfo, function(idx, pInfo) {
-        requests.push(addPlate(pInfo['Plate']));
-      });
+      var requests = [addPlate(gdnaPlate)];
       // We need to wait until all requests are done, so the interface will
       // be ready to be modified
       $.when.apply($, requests).then(function () {
-        $.each(platesInfo, function(idx, pInfo) {
-          var plateId = pInfo['Plate'];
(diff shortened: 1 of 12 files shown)


josenavas:
I forgot to mention that the actual interface behavior has not changed.

josenavas:
@ElDeveloper Can you take a quick look in here? This is blocking some other development and I would like to start moving forward.

ElDeveloper:
On it.

josenavas:
Awesome thanks!

AmandaBirmingham:
Yes, that matches my expectations.  Strictly speaking, each process applies to one or more wells—but in practice so far that generally equates to one process per plate.

josenavas:
Thanks for confirming @AmandaBirmingham !

labman/gui/templates/library_prep_16S.html
```diff
@@ -9,27 +9,35 @@
...
-          $('#sel-primer-' + plateId).val(pInfo['Primer Plate']);
-          $('#epmotion-robot-' + plateId).val(pInfo['EpMotion']);
-          $('#epmotion-tm300-8-tool-' + plateId).val(pInfo['EpMotion TM300']);
-          $('#epmotion-tm50-8-tool-' + plateId).val(pInfo['EpMotion TM50']);
-          $('#mastermix-input-' + plateId).val(pInfo['Master mix']);
-          $('#waterlot-input-' + plateId).val(pInfo['Water lot']);
-        });
-        disableAll();
+        // Need to add a timeout so the interface is ready to set the values
+        setTimeout(function() {
```
ElDeveloper: Is a timeout needed here or would a better way be to wait using:

```javascript
$(function() {
  // things that update the GUI
});
```

Or is the timeout waiting for something else?

josenavas: I don't think that would work. The interface that I'm waiting for here is created in the previous "addPlate" call, which executes an AJAX query, and builds the interface based on the results of the ajax query. This function already gets executed _after_ the ajax called is completed, but it is not ensured that the html that the function create is ready yet.

josenavas:
You're right @tanaes - everything else that was focusing in a single plate was working as expected.

Unfortunately there is a second DB modification that I'm working on right now, so I hope it will be fixed by the time everybody starts working in the morning (except probably @antgonza and @jdereus 😄 ).

In short, the compress gDNA plate information was getting lost (i.e. which plate they originated from). After discussing with @AmandaBirmingham we have decided that the best thing to do is to create a new composition and a new process type to store this information. This is analogous to the normalized gdna process/composition. I've already modified the DB, created the new composition object and fixed the composition tests. I'm working on creating the process object and I will be updating the rest of the code as needed.

labman/gui/templates/library_prep_16S.html
ElDeveloper: I see, thanks for the info!

PR #139, opened 2018-01-25
Creating Compressed gDNA structures

josenavas:
To correctly keep track of the compression process we need to create a new composition. I've modified the DB, created the new composition object, and modified the GDNAPlateCompressionProcess.

There are still a couple of things that I need to modify, but I want to start triggering the tests in travis.


josenavas:
@ElDeveloper @antgonza @tanaes

I have a bug here. It gets exposed when generating the echo pick lists files, but I can't quite find the source of the error. Maybe another set of eyes can help debugging?

josenavas:
I actually found the bug - the only thing left here is adapting @antgonza's prep template code to work with the new structure

labman/db/composition.py
```diff
@@ -685,6 +686,64 @@ def study(self):
...
+        gdna_composition: labman.db.composition.GDNAComposition
+            The source gDNA composition
+
+        Returns
+        -------
+        labman.db.composition.NormalizedGDNAComposition
+            The newly created composition
+        """
+        with sql_connection.TRN as TRN:
+            # Add the tow into the composition table
```
antgonza: tow?

josenavas: Yeah - apparently I have fat fingers 😛

josenavas:
Comments have been addressed and tests pass - I'm going to merge so we can move forward.

Issue #50, opened 2018-01-16
Provide way to see status of samples per Qiita study


adswafford:
Diagram, list, table, etc. 

Could at minimum be a summary, e.g:
1) Select Qiita study
2) Display-
Plated: 500/500 samples 
Extracted: 384/500 samples 
Amplified: 384/500 samples 
Quantified: 384/500 samples 
Pooled: 192/500 samples 
Sequenced: 88/500 samples

Very open to other ideas, so long as it's possible to see where things are in the system somehow.

josenavas:
@qiyunzhu and @mortonjt are working on this.

PR #140, opened 2018-01-25
Study summary

josenavas:
Fixes: https://github.com/jdereus/labman/issues/50

Example view:

screen shot 2018-01-25 at 10 37 50 am

This includes the commits by @mortonjt and @qiyunzhu

### DIFF
--- a/labman/db/tests/test_study.py
+++ b/labman/db/tests/test_study.py
@@ -35,6 +35,17 @@ class TestStudy(LabmanTestCase):
                                   'for Cannabis Soils')
         self.assertEqual(s.creator, User('test@foo.bar'))
         self.assertEqual(s.num_samples, 27)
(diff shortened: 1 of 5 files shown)


josenavas:
@antgonza good point - does that look better now?

labman/db/tests/test_study.py
```diff
@@ -35,6 +35,30 @@ def test_attributes(self):
                                   'for Cannabis Soils')
         self.assertEqual(s.creator, User('test@foo.bar'))
         self.assertEqual(s.num_samples, 27)
+        # self.assertEqual(s.number_samples_plated, 12)
```
antgonza: Code looks much better, do we really need these comments?

josenavas:
ugh tavis... is not starting the build!

PR #141, opened 2018-01-25
Add the ability to copy and paste

### LINKED ISSUE #42: unable to copy/paste values into sheet
try to copy from another spreadsheet it puts all the text into one cell.  need this function since often will receive the samples in a certain order that matches the metadata and it saves a lot of time and prevents errors when can just copy from the metadata instead of typing in each sample.

ElDeveloper:
This PR adds the ability to copy/paste text from Excel (fixes #42) and within the spreadsheet, example:

issue-42

There's a few things to note for reviewers:

- If text is pasted into the cells and the contents exceed the available space, the extra text will be ignored.
- Users can paste from Excel and undo/redo the copy/paste actions, but only the copy paste actions.
- In order to start selecting cells within the table users need to hit ESC, otherwise the selection will be ignored (I've made a note of this in the UI).
- The SlickGrid files have been updated to match the latest tip of www.github.com/6pac/SlickGrid
- I submitted two PRs to 6pac/SlickGrid, because there were two bugs in the source (I've noted this in the License file).
- I made one cosmetic update in plateViewer.js to have the text box be the same size as the containing cell.
- On the first paste action, if there are empty cells these will be shown as "blank" however after the first paste action this no longer happens. I wonder if that's a bug or if there's a problem somewhere.

All the files under `labman/gui/static/vendor/` can be safely ignored as those are direct copies from the originating repo. Perhaps the only file to review there is the license.

### DIFF
--- a/labman/gui/static/js/plateViewer.js
+++ b/labman/gui/static/js/plateViewer.js
@@ -159,7 +159,8 @@ PlateViewer.prototype.initialize = function (rows, cols) {
 
   this.grid = new Slick.Grid(this.target, this.data, sgCols, sgOptions);
 
-  this.grid.setSelectionModel(new Slick.CellSelectionModel());
+  // don't select the active cell, otherwise cell navigation won't work
+  this.grid.setSelectionModel(new Slick.CellSelectionModel({selectActiveCell: false}));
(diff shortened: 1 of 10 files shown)


ElDeveloper:
👍 thanks!

PR #148, opened 2018-01-26
Fix 101

josenavas:
This modifies the DB to store all the information about the normalization and the pooling, as well as modifying the remaining interfaces to load previous processes.

I'm still working on adapting all pages, but I want to kick tests in.


josenavas:
This PR is complete and Fixes #101

PR #149, opened 2018-01-26
ENH: Download preparations for a sequencing run

### LINKED ISSUE #86: Generate PrepInformation
Needs:
- [x] Back-end functionality
- [ ] Modify Sequencing Process interface to make prep information files available for download.

ElDeveloper:
Adds a handler that creates a zip file with all the preparation files
for a given sequencing identifier. Also adds a button in the template
so users can download the data from the sequencing list table.

One thing I couldn't get around was that when running the test suite, the response's code was always 599, regardless of the URI (even endpoints that already existed). I couldn't test much, only via the UI. It would be good if someone can shed some light as to what else I can do, or if I'm missing something (maybe because of my local installation).

Fixes #86

### DIFF
--- a/labman/gui/handlers/process_handlers/__init__.py
+++ b/labman/gui/handlers/process_handlers/__init__.py
@@ -19,7 +19,8 @@ from .pooling_process import (
     PoolPoolProcessHandler, LibraryPoolProcessHandler,
     ComputeLibraryPoolValueslHandler, DownloadPoolFileHandler)
 from .sequencing_process import (
-    SequencingProcessHandler, DownloadSampleSheetHandler)
+    SequencingProcessHandler, DownloadSampleSheetHandler,
+    DownloadPreparationSheetsHandler)
 from .normalization_process import (
     NormalizationProcessHandler, DownloadNormalizationProcessHandler)
 from .primer_working_plate_creation_process import (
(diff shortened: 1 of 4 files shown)


josenavas:
@ElDeveloper there are a couple of small flake8 errors here :)

josenavas:
BTW, 599 is request timeout - so I think the problem is that generating the zip file is taking longer than expected.

You can probably fix it by using this: https://github.com/tornadoweb/tornado/issues/1753

ElDeveloper:
Will fix the style problems for sure. As for the 599, this was happening even for tests I didn’t write. I also noticed that other tests don’t check for 200 return codes. Any ideas?

Yoshiki Vázquez-Baeza

josenavas:
Can you actually check the contents? i.e. the contents show as you expected?

ElDeveloper:
No, the response seems to be malformed from the test suite but it works as expected from the UI ¯\_(ツ)_/¯ 

Yoshiki Vázquez-Baeza

josenavas:
That's so weird - can you actually put a check for the contents and let travis run the suite? I'm wondering if it has to do with your set up.

ElDeveloper:
Thanks for the help @josenavas tests are passing now 👍

PR #155, opened 2018-01-26
Activate a plate if none are added

### LINKED ISSUE #115: Autoselect the first study in the plating page

wasade:
fixes #115 i think


labman/gui/static/js/plate.js
```diff
@@ -55,6 +55,11 @@ function add_study(studyId) {
     var $buttonElem = $("<button class='btn btn-danger btn-circle pull-right' onclick='remove_study(" + studyId + ");'>");
     $buttonElem.append("<span class='glyphicon glyphicon-remove'></span>")
     $aElem.append($buttonElem);
+
+    if($('#study-list').children().length === 0) {
```
josenavas: sorry @wasade Reviewing the code I realized that this call should be done after the append, otherwise I don't think the activate_study would work as expected. Can you move this block after the append in the next line and change the if condition to be `=== 1`? Sorry for not realizing this while we're talking!

PR #169, opened 2018-01-30
Add a frozen column for the row names.

ElDeveloper:
@josenavas, I ended up having to correct some of the indices. Turns out it is not very easy to hide a column from the data without deleting the information altogether. Please double check the changes I've made to "uncorrect" indices, I don't have a good way to validate that these are indeed accurate.

### DIFF
--- a/labman/gui/static/js/plateViewer.js
+++ b/labman/gui/static/js/plateViewer.js
@@ -18,6 +18,11 @@ function PlateViewer(target, plateId, processId, rows, cols) {
 
   this.container = $('#' + target);
 
+  /*
+   * HACK: SlickGrid doesn't currently support frozen columns hence we are
+   * using two grids to make it look like the first column is frozen. Once
+   * the feature makes it into the SlickGrid repo, we can remove this
+   */
   this.target = $('<div name="main-grid"></div>');
   this._frozenColumnTarget = $('<div name="frozen-column" ' +
                                'class="spreadsheet-frozen-column"></div>');
--- a/labman/gui/static/js/plateViewer.js
+++ b/labman/gui/static/js/plateViewer.js
@@ -14,10 +14,29 @@
  *
  **/
 function PlateViewer(target, plateId, processId, rows, cols) {
-  this.target = $('#' + target);
+  var height = '250px';
+
+  this.container = $('#' + target);
+
+  this.target = $('<div name="main-grid"></div>');
+  this._frozenColumnTarget = $('<div name="frozen-column" ' +
+                               'class="spreadsheet-frozen-column"></div>');
+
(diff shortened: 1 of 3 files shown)


labman/gui/static/js/plateViewer.js
```diff
@@ -14,10 +14,34 @@
...
 function PlateViewer(target, plateId, processId, rows, cols) {
-  this.target = $('#' + target);
+  var height = '250px';
+
+  this.container = $('#' + target);
+
+  /*
+   * HACK: SlickGrid doesn't currently support frozen columns hence we are
+   * using two grids to make it look like the first column is frozen. Once
+   * the feature makes it into the SlickGrid repo, we can remove this
```
wasade: Is there an associated issue with SlickGrid that can be referenced by chance?

labman/gui/static/js/plateViewer.js
```diff
@@ -221,9 +263,7 @@ PlateViewer.prototype.initialize = function (rows, cols) {
       return;
     }
     var row = $(this).data("row");
-    // col - 1 to get the correct index, since column 0 is the header
-    // in the actual slickgrid
-    var col = $(this).data("col") - 1;
+    var col = $(this).data("col");
```
wasade: The -1 isn't needed anymore?

labman/gui/static/js/plateViewer.js
```diff
@@ -413,7 +452,7 @@ PlateViewer.prototype.updateWellCommentsArea = function () {
   var wellId;
   for (var i = 0; i < that.rows; i++) {
     for (var j = 0; j < that.cols; j++) {
-      wellId = rowId + (j + 1);
+      wellId = rowId + j;
```
wasade: ...oh, the -1 above was just redundant?

labman/gui/static/js/plateViewer.js
ElDeveloper: The indices refer to the first column, the one that used to have the row names (A, B, C ...). Since the row names have now been moved to a new table the `- 1` is no longer needed.

labman/gui/static/js/plateViewer.js
```diff
@@ -21,7 +21,8 @@ function PlateViewer(target, plateId, processId, rows, cols) {
   /*
    * HACK: SlickGrid doesn't currently support frozen columns hence we are
    * using two grids to make it look like the first column is frozen. Once
-   * the feature makes it into the SlickGrid repo, we can remove this
+   * the feature makes it into the SlickGrid repo, we can remove this. See
+   * this GitHub issue: https://github.com/6pac/SlickGrid/issues/26
```
wasade: 👍

PR #184, opened 2018-02-02
Fixes 172 and 152

josenavas:
Fixes #172 
Fixes #152 

I want to get the tests kicked off - I'm working on adding an interface to let a user give access permissions as well as an interface to create controls.

PR #188, opened 2018-02-05
Fixes 174

### LINKED ISSUE #174: Plate add interface very slow
Dialogues that present a list of plates can be extremely slow (e.g. the "add plates" modal on the DNA Quant interface takes ~7 seconds to populate with the test db w/ 5 plates). 

The mechanism for generating these tables needs to be much faster.

josenavas:
Depends on https://github.com/jdereus/labman/pull/184 so review/merge that one first.

Fixes #174


ElDeveloper:
@josenavas would you mind updating this PR to resolve conflicts?

josenavas:
@ElDeveloper @tanaes are you available to take a look at this?

tanaes:
Taking a look now

ElDeveloper:
Also having a look at it.

labman/db/plate.py
```diff
@@ -219,16 +223,21 @@ def list_plates(plate_types=None, only_quantified=False,
             if only_quantified:
                 sql_join = ("JOIN qiita.concentration_calculation "
```
tanaes: Now that we're adding `sql_studies`, is `sql_join` the right name here? (just asking for clarity/future readability)

labman/db/support_files/db_patch_manual.sql
```diff
@@ -3525,3 +3525,81 @@ INSERT INTO qiita.shotgun_combo_primer_set (shotgun_primer_set_id, i5_primer_set
...
+                    JOIN qiita.sample_composition USING (sample_id)
+                    JOIN qiita.composition USING (composition_id)
+                    JOIN qiita.well USING (container_id)
+                WHERE plate_id = in_plate_id);
+    ELSIF plate_type = 'gDNA' THEN
+        RETURN (SELECT array_agg(DISTINCT study_title)
+                FROM qiita.study
+                    JOIN qiita.study_sample USING (study_id)
+                    JOIN qiita.sample_composition USING (sample_id)
+                    JOIN qiita.gdna_composition gdna USING (sample_composition_id)
```
tanaes: Seems unfortunate that you need to have this big series of `ELSIF` statements. I assume it's because these trace back to unique tables per `plate_type`?

josenavas:
Thanks!

labman/db/plate.py
josenavas: I think it is still accurate, since it is adding more tables to the JOIN part of the query

tanaes:
Nice:
slow

fast

labman/db/support_files/db_patch_manual.sql
josenavas: Yes, It could've been done by creating functions for each plate type, but that would result on multiple queries that, although they're all happening at once in the DB server, this way is still faster cause it is specialized for each plate type, performing a single query per plate rather than multiple queries per plate to perform the traceback.

josenavas:
💥

tanaes:
@josenavas ready for me to merge?

josenavas:
Yup!!

PR #192, opened 2018-02-08
Adding integration tests

josenavas:
@wasade This adds integration tests to the test suite - can you take a look to see if this is what you would expect?

Thanks!


labman/db/tests/integration_tests.py
```diff
@@ -0,0 +1,214 @@
...
+    LibraryPrep16SProcess, NormalizationProcess, QuantificationProcess,
+    LibraryPrepShotgunProcess, PoolingProcess, SequencingProcess)
+from labman.db.user import User
+from labman.db.plate import PlateConfiguration, Plate
+from labman.db.equipment import Equipment
+from labman.db.composition import ReagentComposition
+from labman.db.sql_connection import TRN
+
+
+def get_samples():
```
tanaes: Is this... *all* of the samples in the qiita db?

labman/db/tests/integration_tests.py
```diff
@@ -0,0 +1,214 @@
...
+    kappa = ReagentComposition(4)
+    stub = ReagentComposition(5)
+    shotgun_process = LibraryPrepShotgunProcess.create(
+        user, norm_plate, 'Test Shotgun Library %s' % datetime.now(), kappa,
+        stub, 4000, Plate(19), Plate(20))
+    shotgun_plate = shotgun_process.plates[0]
+    return shotgun_process, shotgun_plate
+
+
+def create_plate_pool_process(user, shotgun_quant_process, shotgun_plate,
```
tanaes: I think this is generic for shotgun and amplicon pools, right? so maybe just `quant_process` and `plate`?

labman/db/tests/integration_tests.py
```diff
@@ -0,0 +1,214 @@
...
+    # Create sequencing pool process
+    seq_pool_process = create_pools_pool_process(
+        user, pool_quant_process, [plate_pool_process.pool])
+    # Sequencing process
+    create_sequencing_process(user, [seq_pool_process.pool])
+
+
+def shotgun_workflow(user, samples):
+    # Sample Plating
+    sp_process, sample_plate = create_sample_plate_process(user, samples[:96])
```
tanaes: Typically these gDNA runs have four plates... But maybe not worth doing that for this integration test, since it's just exercising each of the steps rather than trying to actually load the system?

labman/db/tests/integration_tests.py
josenavas: Yes, even on the test environment (with more than 500k samples), it takes 1-2 seconds to run.

labman/db/tests/integration_tests.py
josenavas: That's correct - although I do want to extend the functionality of this functions to be able to handle a parameter with the number of plates to create.

labman/db/tests/integration_tests.py
josenavas: Good point - I was copy/pasting code around :D

labman/db/tests/integration_tests.py
```diff
@@ -0,0 +1,213 @@
...
+    create_sequencing_process(user, [pool_process.pool])
+
+
+@click.command()
+def integration_tests():
+    samples = get_samples()
+    user = User('test@foo.bar')
+    amplicon_workflow(user, samples)
+    shotgun_workflow(user, samples)
+
```
wasade: would it be possible to assert the sample sheets contain the expected values?

josenavas: Done - I had to modify the code to ensure the sorting of the samples in the sample sheet.

labman/db/tests/integration_tests.py
```diff
@@ -0,0 +1,375 @@
...
+\[Reads\]
+151
+151
+
+\[Settings\]
+ReverseComplement,0
+
+\[Data\]
+Lane,Sample_ID,Sample_Name,Sample_Plate,Sample_Well,I7_Index_ID,index,I5_Index_ID,index2,Sample_Project,Description
```
tanaes: This looks ok, and should work as long as the db is fresh -- otherwise, possible that the sample:index associations will not match

josenavas: That's correct  - and expected. If we want to ensure associations and results this is the only way of doing it.

tanaes: Sweet!

Issue #112, opened 2018-01-23
Add grid showing sample ids on pooling


josenavas:

ElDeveloper:
Can someone provide a more in-depth explanation of this issue?

josenavas:
@ElDeveloper, when the user has computed the values for pooling, this is what is shown to the user:

<img width="1215" alt="screen shot 2018-02-02 at 2 27 13 pm" src="https://user-images.githubusercontent.com/2501478/35757719-459c7492-0825-11e8-88cf-22a084d3815b.png">

The goal is to have another grid above or below that shows the sample plate layout (i.e. the layout when the user goes to 'Plate samples'). This way they can match the pooling values with the samples. Does that make sense?

ElDeveloper:
Thanks @josenavas, this helps a lot!

josenavas:
👍

Issue #69, opened 2018-01-18
figure generation and return/display


tanaes:
There are a lot of steps that would benefit from / absolutely need some sort of visual feedback to illustrate the success of particular steps. 

For example, see the plate concentration visualization images from the current shotgun pipeline [Jupyter notebooks](https://github.com/tanaes/metagenomics_pooling_notebook/blob/master/metagenomics_pipeline_MiniPiconorm.ipynb):

<img width="626" alt="screenshot" src="https://user-images.githubusercontent.com/8292608/35077467-69036be0-fbb2-11e7-83d8-a0d2de760358.png">

These plate visualizations should be displayed when uploading anything that quantifies DNA per well of a plate, providing a visual check to indicate that things worked—for example, on the Plate Quantification interface, or the Library Pooling interface. 

The code for this particular figure generation is [already written](https://github.com/tanaes/metagenomics_pooling_notebook/blob/f3f9e6add8472df9663adb9bb995da238c550a5e/metapool/metapool.py#L489). We just need a way to deploy and integrate into the HTML / Javascript interface.

ElDeveloper:
As per @josenavas' and @tanaes's suggestion, this issue ties in very well with #112. @nreeve17 and I are working on this to create an interactive heatmap that displays the quantities and sample identifiers.

PR #193, opened 2018-02-10
Adding heatmap and histogram to pooling plates screen

nreeve17:
This addresses issue #69 and issue #112. This PR also includes Plotly as a dependency. @ElDeveloper and I pair-programmed portions of this.

Here's a demo of the new functionality:

heatmap


nreeve17:
I forgot to mention that the tests for the function we modified do not explicitly test the function itself, but rather just that it outputs the right variables. It might be good to have more explicit testing. Looks like the travis build failed, I will look into fixing that

ElDeveloper:
Looks like the problems are mostly stylistic in the Python files:

test/test_pooling_process.py:123:57: E502 the backslash is redundant between brackets

labman/gui/handlers/process_handlers/test/test_pooling_process.py:124:39: E127 continuation line over-indented for visual indent


Yoshiki Vázquez-Baeza

labman/db/plate.py
```diff
@@ -321,6 +321,7 @@ def layout(self):
         with sql_connection.TRN as TRN:
             pc = self.plate_configuration
             layout = []
+    # return euclidean(x.values[:-1], vector.loc[x.grouping].values)
```
josenavas: Can you remove the commented code?

labman/gui/handlers/process_handlers/pooling_process.py
```diff
@@ -114,13 +115,29 @@ def make_2D_arrays(plate, quant_process):
...
+    plate_names = np.empty_like(layout, dtype='object')
     for comp, raw_conc, conc in quant_process.concentrations:
         well = comp.container
         row = well.row - 1
         column = well.column - 1
         raw_concs[row][column] = raw_conc
         comp_concs[row][column] = conc
-    return raw_concs, comp_concs
+
+        if isinstance(comp, LibraryPrep16SComposition):
```
josenavas: Both here and in the other if below, can you create a variable to store the 'sample_composition' element? Since you have to access that twice, recovering it once and accessing its properties would be faster, since you're doing less queries to the DB (you will do (n/2) + 1 calls instead of n)

labman/gui/templates/library_pooling.html
```diff
@@ -33,8 +35,11 @@
           $.each(poolParams[funcData['function']], function(idx, paramElem) {
             parameterName = paramElem['prefix'] + plateId;
             $('#' + parameterName).val(funcData['parameters'][paramElem['prefix']]);
-            // Create the grid
-            createPoolingGrid(plateId, {% raw pool_values %});
```
josenavas: Is the `createPoolingGrid` function used anywhere? If not, can it be removed?

labman/gui/templates/library_pooling.html
```diff
@@ -59,6 +64,132 @@
...
 
+  function createHeatmap(plateId, concentrations, blanks, names) {
+    var container = $('#pool-results-' + plateId);
+    var heatmap = $('<div id="heatmap-' + plateId + '"></div>');
+    var histogram = $('<div id="histogram-' + plateId + '"></div>');
+
+    container.append(heatmap);
+    container.append(histogram);
+
+    blanks = blanks.reverse();
```
josenavas: Can you add a comment explaining why these needs to be reversed?

labman/gui/handlers/process_handlers/pooling_process.py
ElDeveloper: 👍 For sure, totally missed this!

ElDeveloper:
Tests are passing in #194, closing in favor of that PR.

PR #194, opened 2018-02-12
Plotting functionality

ElDeveloper:
Supersedes #193. It addresses the review comments by @josenavas (see the commits after 
6ee5aa1).


ElDeveloper:
I've also added a grid to the heatmap to aid in legibility:

screen shot 2018-02-12 at 11 26 17 am

tanaes:
This plot looks great, was just exploring with @ElDeveloper and submitted an issue to extend functionality to other interface page. 

Only minor comment is that we may require the ability to do some clipping or transformation for some uses. Examples include quantification, where there sometimes some values that are very high and drown out the median values on the viridis colormap.

PR #196, opened 2018-02-12
Allow users to download all the plate maps of a study

### LINKED ISSUE #190: export all plate maps from a single study to a tsv/excel file
Requested by Lindsay:

On the study page, it would be useful to be able to download all the plate maps of a single study in a excel-compatible file.

josenavas:
Fixes #190

I first implemented it using a function on the plate object, but after a test with a study with 200 plates I noticed it was a bit slow (see below) so I change the implementation to return all the information at once, although the code is a bit more complex. I've added comments that explain what I'm doing on the code so it is easier to maintain. As seen below, the current implementation is fast even with a study with 200 plates. Also below, a GIF with the changes in the GUI.

```python
In [1]: from labman.db.study import Study

In [2]: def f1(sid):
   ...:     study = Study(sid)
   ...:     res = [p.get_plate_map() for p in study.sample_plates]
   ...:     return '\n\n'.join(res)
   ...: 
   ...: def f2(sid):
   ...:     return Study(sid).generate_sample_plate_maps()
   ...: 

In [3]: %timeit t = f1(1)
12 s ± 330 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [4]: %timeit t = f2(1)
519 ms ± 8.84 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

platemaps

### DIFF
--- a/labman/db/study.py
+++ b/labman/db/study.py
@@ -6,9 +6,12 @@
 # The full license is in the file LICENSE, distributed with this software.
 # ----------------------------------------------------------------------------
 
+import numpy as np
+
 from . import base
 from . import sql_connection
 from . import user
+from . import container as container_module
 
 
 class Study(base.LabmanObject):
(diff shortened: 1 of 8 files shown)


labman/db/study.py
```diff
@@ -64,6 +67,86 @@ def creator(self):
         """The user that created the study"""
         return user.User(self._get_attr('email'))
 
+    def generate_sample_plate_maps(self):
+        """Generates a string with the plate maps in a csv format
+
+        Returns
+        -------
+        str
```
ElDeveloper: This seems to be incomplete.

labman/db/study.py
```diff
@@ -64,6 +67,86 @@ def creator(self):
...
+                     ORDER BY plate_id"""
+            TRN.add(sql, [self.id])
+            res = TRN.execute_fetchindex()
+            # Create a list of numpy arrays to store the layouts
+            layouts = [np.zeros((nr, nc), dtype=object)
+                       for _, _, nr, nc, _ in res]
+            # Loop through all the plates and populate the layouts with the
+            # contents of the well
+            for idx, (_, _, _, _, layout) in enumerate(res):
+                # Description of the string mangling: The `array_agg` operation
```
ElDeveloper: Thanks for adding this description, would have been pretty hard to follow otherwise. If you still think eval might be a way to go (security people might cringe about that), perhaps replacing the curly braces with at the beginning and end and replacing them for open and close squared braces would do the trick. Alternatively, if this is something that's been needed elsewhere it might also make sense to leave this parsing code as is but centralize it in a function outside of this method. That way the function can be properly tested for edge cases, documented, etc.

charles-cowart:
Looked at resolving the conflicts, and there appears to be substantial changes made to study.py since the time this mod was made. I'm willing to close the PR without merging, but I'd like a second opinion from either @ElDeveloper or @AmandaBirmingham, if possible.

wasade:
If easier to rewrite then +1. It is most likely the case that some of the logic here can be harvested

AmandaBirmingham:
I strongly suspect easier to rewrite than to merge after this long a time.  Vote to close without merge but keep a link to this PR in the (still open) issue #190 so whoever gets that task can harvest logic from here.

wasade:
+1

PR #197, opened 2018-02-13
Adding script to easily add data to the DB

josenavas:
I've renamed the "integration_tests.py" script to "tester.py" and I've modified it to have another command that allows to add multiple data to the DB. Since the script can take a while to run (depending on the number of plates per workflow to create), the script prints the current progress so the user has some feedback:

```sh
$ labman/db/tests/tester.py stress_tests --num_plates 20
Amplicon workflow
	Sample plating process....................Ok
	gDNA extraction process....................Ok
	Amplicon library prep process....................Ok
	Amplicon library quantification process....................Ok
	Plate pooling process....................Ok
	Plate pool quantification process...Ok
	Sequencing pool process...Ok
	Sequencing process...Ok
Shotgun workflow
	Sample plating process....................Ok
	gDNA extraction process....................Ok
	gDNA compression process.....Ok
	compressed gDNA quantification process.....Ok
	gDNA normalization process.....Ok
	Shotgun library prep process.....Ok
	Shotgun library quantification process.....Ok
	Pooling process.....Ok
	Sequencing process.....Ok
```

### DIFF
--- a/labman/db/tests/tester.py
+++ b/labman/db/tests/tester.py
@@ -211,6 +211,8 @@ def integration_tests_shotgun_workflow(user, samples):
 
 @tester.command()
 def integration_tests():
+    """Creates one amplicon and one shotgun workflow from plating to sequencing
+    """
     samples = get_samples()
     user = User('test@foo.bar')
     amplicon_seq_process = integration_tests_amplicon_workflow(user, samples)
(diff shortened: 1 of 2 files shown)


labman/db/tests/tester.py
```diff
@@ -224,6 +229,175 @@ def integration_tests():
...
+                 'parameters': {"dna_amount": 240, "min_val": 1, "max_val": 15,
+                                "blank_volume": 2, "robot": 6,
+                                "destination": 1}}))
+        print('.', end='', flush=True)
+    print('Ok', flush=True)
+
+    # Quantify pools
+    print('\tPlate pool quantification process', end='', flush=True)
+    pool_quant_processes = []
+    for pos in range(0, num_plates, 8):
```
tanaes: It looks like this is trying to pool sets of 8 amplicon pools per sequencing process. Should the logic be more like on line 333, where you're adding in just the set of 8 pools in this increment?

labman/db/tests/tester.py
```diff
@@ -224,6 +229,175 @@ def integration_tests():
...
+        print('.', end='', flush=True)
+    print('Ok', flush=True)
+
+
+@tester.command()
+@click.option('--num_plates', required=False, type=click.IntRange(1, None),
+              default=10, show_default=True,
+              help='Number of plates to create per workflow')
+def stress_tests(num_plates):
+    samples = get_samples()
```
tanaes: So just as a refresher, this is going to grab each of the samples that currently exist in the Qiita db, and add them sequentially to plates -- maintaining study associations? Since there's only one study currently in my test db it looks like it's just cycling through those samples.

josenavas: This is getting the list of samples in the entire DB. The stress_test command loops over all the samples in the DB. The sample id is linked to a study in a table described in the Qiita repository `qiita.study_sample`, and the table `qiita.sample_composition` has a FK to that table (and hence maintaining the relation study-sample). Since we want to be able to add, let's say, 100 plates in the DB (I did this on my DB to stress test my previous PR) the `stress_tests` function loops over all the samples in the DB. If we run this in the maptest system, it will be looping through all the samples in the DB.

labman/db/tests/tester.py
josenavas: The pool quantification process is done in batch, rather than having a 1-1 relationship (matches the nature of the process in the wet lab) so it is grouping it here and providing that information to the `create_pools_pool_process` function.

ElDeveloper:
This looks good, just about to try it out. Perhaps my only suggestion would be to use click's progressbar instead of printing dots. But this is rather minor. Link, just in case:

http://click.pocoo.org/5/utils/#showing-progress-bars

tanaes:
I asked @eldeveloper if these were the progress bars that kill my /
travis's log files and we verified that they are not, and are instead good.
Click is good.

-j

ElDeveloper:
One other thing, now that the script is a click command, would be to include a short docstring describing what `integration_tests` does and what `stress_tests` does.

labman/db/tests/tester.py
tanaes: but looking at the logic, isn't the case that `pools` here is going to be the entire set of N plates in `plate_pool_processes` for each `pool_quant_process`, when really it should just be the next set of 8?

josenavas: Nope, this for loop is grouping the pools in sets of 8 to create a single sequencing pool. Then each of this individual sequencing pools are individually "sequenced".

josenavas: I just talked with @ElDeveloper and I saw the bug on the code - I will push the fix soon.

ElDeveloper:
👍

PR #198, opened 2018-02-14
Interactivity improvements

ElDeveloper:
The new functionality comes from talking to @tanaes. When visualizing concentrations in a plate (as represented by a heatmap), it is useful to change the range of the colorscale, in order to do this, we've added a slider to manipulate this value, and set some appropriate defaults depending on the plate types. This will also be useful as we reuse this component for other steps in the processing pipeline.

Demo of the new functionality:
slidere

I've also cleaned up some of the JS code to avoid multiple traversals of the main arrays.

### DIFF
--- a/labman/gui/templates/library_pooling.html
+++ b/labman/gui/templates/library_pooling.html
@@ -22,6 +22,8 @@
       bootstrapAlert('Loading data...', 'info');
       var funcData = {% raw pool_func_data %};
       var plateType = funcData['function'] === 'amplicon' ? '16S library prep' : 'shotgun library prep';
+      var defaultClipping = clippingForPlateType(plateType);
+
       $('#plate-type-select').val(plateType);
       $('#plate-list-div').show();
       var requests = [addPlate({{input_plate}})];
@@ -38,7 +40,8 @@
 
             // create the visualizations
             createHeatmap(plateId, {% raw pool_values %},
-                          {% raw pool_blanks %}, {% raw plate_names %}, 30);
+                          {% raw pool_blanks %}, {% raw plate_names %},
+                          defaultClipping);
 
             // Add the download button
             $('#download-' + plateId).find('a').attr('href', '/process/poollibraries/' + processId + '/pool_file');
@@ -176,7 +179,7 @@
         xgap: 1,
         ygap: 1,
         zmin: minConcentration,
-        zmax: 30
+        zmax: defaultClipping
       }
     ];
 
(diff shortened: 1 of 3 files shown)


labman/gui/templates/library_pooling.html
```diff
@@ -193,21 +221,59 @@
...
   };
 
+  /**
+   * Determine an appropriate clipping value for the heatmap's maximum value.
+   *
+   * @param {Float} plateType String describing the plate type.
+   * @returns A maximum expected value for the colorscale of a heatmap.
+   *
+   */
+  function clippingForPlateType(plateType) {
```
ElDeveloper: @tanaes I am sure 💯 % that I got these wrong (forgot to write the numbers down), I would appreciate if you can remind me the correct values.

josenavas:
Code looks good 👍

labman/gui/templates/library_pooling.html
```diff
@@ -193,21 +221,59 @@
...
+    if (plateType == '16S library prep') {
+      return 100;
+    }
+    else {
```
tanaes: maybe make this an `elif (plateType == 'shotgun library prep')` statement to explicitly address the possibility of other plate types?

These are pooling values, so correct defaults are 1000 for shotgun and 15 for 16S. 

IF using slider to clip, do we also need to set the slider range values?

labman/gui/templates/library_pooling.html
```diff
@@ -193,21 +221,59 @@
...
       yaxis: {title: "Number Samples"},
       title: 'DNA Concentration Distribution'
     };
-    var histdata = [trace, blank_trace];
+    var histogramData = [nonBlankData, blankData];
+
+    Plotly.plot($heatmap.attr('id'), heatmapData, heatmapLayout);
+    Plotly.plot($histogram.attr('id'), histogramData, histogramLayout);
+
+    // setup some interactivity to clip the values in the histogram
```
tanaes: Does this want to be simplified into a text box? The advantage is that it wouldn't require any default max values.

labman/gui/templates/library_pooling.html
```diff
@@ -75,15 +78,41 @@
...
+   * @param {Float} defaultClipping A number representing the default clipping
+   * value for the heatmap. Depends on the processing stage and the type of
+   * data being processed.
    *
    */
-  function createHeatmap(plateId, concentrations, blanks, names) {
-    var container = $('#pool-results-' + plateId);
-    var heatmap = $('<div id="heatmap-' + plateId + '"></div>');
-    var histogram = $('<div id="histogram-' + plateId + '"></div>');
+  function createHeatmap(plateId, concentrations, blanks, names,
```
tanaes: We want to also have this code in the quantify interface. 

Should we be locating this function in a different, centralized file and importing?

ElDeveloper:
I've addressed the changes, mainly changing the slider for a spinner:

spinner

tanaes:
niiice

Issue #195, opened 2018-02-12
Fixes to quantify plates interface


tanaes:
Noticed some minor issues with quantification while looking at heatmaps with @ElDeveloper. Currently, you can select a plate that has already been quantified and go through all the steps of quantification without any notification that it has already been done. The uploaded values appear to be silently discarded.

I suggest the following changes to this interface:

- [ ] update logic so that you can (optionally) re-quantify a plate; for example if the first quant was bad, or done a long time ago

- [ ] update interface so that if a plate has already been quantified, the 'upload quant file' button is disabled until you click something to enable overwriting the quantification values

- [ ] update interface so that if a plate has already been quantified, OR you have just uploaded quantification values, it displays the quantification heatmap / histogram

ElDeveloper:
I am trying to add the necessary bits to display a quantified plate as a heatmap + histogram (improvement 3), but I am having a hard time figuring out the sample names of each of the wells. Seems unclear to me how to do this as the `QuantificationProcess` doesn't exist until after the plate is saved. Are the visuals only to be displayed after the process has been saved? My intuition says that it would be useful to *see* the data before saving it, but I don't really know the process.

tanaes:
Hmm, yes... Visualization happens pre-saving as a safeguard (or, ideally,
we'd be able to go back to a plate to visualize of course).

I think you'd have to retrieve the sample names as an array from the plate
you were going to quantify and just associate that with the same cells on
the file you read in, rather than pulling them from the
`QuantificationProcess`.

ElDeveloper:
Thanks for the pointers @tanaes, that clears things up.

AmandaBirmingham:
@tanaes could you please provide a sample plate reader output file that I could use while working on this issue?

tanaes:
This should work: 10-13-17_FinRisk_33-36_library_quant.txt
<https://github.com/tanaes/metagenomics_pooling_notebook/blob/master/test_data/Quant/MiniPico/10-13-17_FinRisk_33-36_library_quant.txt>

AmandaBirmingham:
@tanaes Assuming we are talking about the behavior of Plates > PicoGreen Quantification, let me make an important correction to the above issue description: the uploaded values are NOT discarded.  The database supports multiple quantifications for any given well/plate/etc, so every time the user chooses to save a quantification result, it goes into the database as a distinct quantification_process with its own entirely distinct set of entries in the concentration_calculation table. 

Given this, here's my input on the three proposed changes:

[...]

PR #201, opened 2018-02-15
ENH: Add a heatmap display for the quantification step

ElDeveloper:
The quantification step ditches the grid view and now reuses the display
element for the pooling step:

quantification

This PR also parametrizes the colormap to fit the step, by recommendation of @tanaes Viridis is used for quantification.

Related to #195

### DIFF
--- a/labman/gui/handlers/process_handlers/quantification_process.py
+++ b/labman/gui/handlers/process_handlers/quantification_process.py
@@ -15,7 +15,8 @@ from labman.gui.handlers.base import BaseHandler
 from labman.db.plate import Plate
 from labman.db.process import QuantificationProcess
 from labman.db.composition import (LibraryPrepShotgunComposition,
-                                   LibraryPrep16SComposition)
+                                   LibraryPrep16SComposition, GDNAComposition,
+                                   CompressedGDNAComposition)
 
 
 class QuantificationProcessParseHandler(BaseHandler):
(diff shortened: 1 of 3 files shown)


labman/gui/handlers/process_handlers/quantification_process.py
```diff
@@ -37,9 +39,43 @@ def post(self):
             pc = plate.plate_configuration
             concentrations = QuantificationProcess.parse(
                 file_content, rows=pc.num_rows, cols=pc.num_columns)
+
+            names = np.empty_like(plate.layout, dtype='object')
+            blanks = np.zeros_like(plate.layout, dtype=bool)
+
+            for i, full_row in enumerate(plate.layout):
```
tanaes: For future use, can we add a comment line to this block to clarify its purpose? (which I think here is just to get the names and blank status, right?)

labman/gui/templates/quantification.html
```diff
@@ -28,33 +24,36 @@
...
 
+  /**
+   * Determine an appropriate clipping value for the heatmap's maximum value.
+   *
+   * @param {Float} plateType String describing the plate type.
+   * @returns A maximum expected value for the colorscale of a heatmap.
+   *
+   */
+  function clippingForPlateType(plateType) {
+    if (plateType == '16S library prep') {
```
tanaes: These should be updated to include:

`['gDNA', 'compressed gDNA']`

with default clipping at 20

ElDeveloper:
@tanaes great suggestions, they've now been addressed. 👍

labman/gui/handlers/process_handlers/quantification_process.py
```diff
@@ -37,9 +39,45 @@ def post(self):
...
+                for j, well in enumerate(full_row):
+
+                    # some wells have no compositions at all so skip those
+                    if well is None:
+                        continue
+                    comp = well.composition
+
+                    # cache the sample compositions to avoid extra intermediate
+                    # queries
+                    if isinstance(comp, LibraryPrep16SComposition):
```
josenavas: I think you also need two more types here: `GDNAComposition` and `CompressedGDNAComposition`, since those two types can also be quantified.
To get the sample composition from a `GDNAComposition`:
```python
comp.sample_composition
```
from a `CompressedGDNAComposition`:
```python
comp.gdna_composition.sample_composition
```

labman/gui/handlers/process_handlers/quantification_process.py
```diff
@@ -55,22 +56,22 @@ def post(self):
 
                     # cache the sample compositions to avoid extra intermediate
                     # queries
-                    if isinstance(comp, LibraryPrep16SComposition):
+                    if isinstance(comp, (LibraryPrep16SComposition,
```
josenavas: I think I may have misinformed you when I give you the information from the GDNAComposition. It should be:
```
comp.sample_composition
```

tanaes:
@josenavas can we merge this?

Issue #203, opened 2018-02-16
Fix amplicon quantification and pooling


tanaes:
Currently, amplicon quantification and pooling doesn't work correctly. 

The way the wet lab does amplicon pooling currently is that it:
1. calculates DNA concentrations
2. calculates the volume that would be necessary to pool 240 ng of that sample
3. if the required volume exceeds 15 µL and the sample is not a blank, pool 2 µL 
4. pool 2 µL from the two highest-concentration blanks

But what's happening in the code currently is that the first three elements of this are (sort of) happening *in the concentration calculation*: 

https://github.com/jdereus/labman/blob/7a4df3170e6a511336fdb7294521999b34d3df77/labman/db/process.py#L1825-L1833

This means that rather than actual estimated concentrations being stored in the DB, we are storing some actual concentrations (high-conc samples) and some fake concentrations (low-conc samples and blanks), then doing a trivial calculation in the pooling method:

https://github.com/jdereus/labman/blob/7a4df3170e6a511336fdb7294521999b34d3df77/labman/db/process.py#L1866-L1881

Additionally, Greg requests some ability so selectively choose the number of blanks to pool (e.g in a plate with lots of high-conc blanks, might want to sequence several to see what's going on).

...

I suggest the following:

1. Consolidate the concentration logic for 16S and shotgun libraries into a single method. They're fundamentally the same: just estimating molarity based on DNA concentration and average fragment size. There's no reason to separate them.
2. Consolidate the pooling logic for 16S and shotgun libraries. The 16S method is identical to the 'min-vol' method for shotgun, with the exception that blanks are treated specially. No reason not to use the same methodology here and just pass parameters to affect behavior. The parameters should be defined by the js interface.
3. Expose the blank-selection criteria in the interface, and modify heatmap plots such that included blanks are visually differentiated (changing emojis for those wells?)

tanaes:
Closed via #222

PR #208, opened 2018-02-21
WIP: Fix amplicon pooling

tanaes:
This is a first stab at fixing issue #203. 

I've modified the underlying functional logic and HTML code to enable sensible application of minvol and eqvol pooling to both amplicon and shotgun library plates. This also involves passing per-library-type defaults and making some different choices when calculating pooling (for example, amplicon is pooling by raw ng/uL concentrations and shotgun by calculated nM concentrations)

Still to do:

- [ ] Add blank-specific behavior; for example, allow user to pool only N blanks at X volume.
- [ ] update heatmap display so that it's got the right titles (e.g. it's displaying volumes, not concentrations)
- [ ] improve interface -- not optimal allocation of space to e.g. parameters entry boxes vs descriptions

Some specific questions for review:
1. There are parameter values that are stored (e.g. robot used and destination) but that aren't used by the pooling calculation itself. Are those being handled the correct way? E.g. labman/gui/handlers/process_handlers/pooling_process.py L30
2. some of these parameters might be more appropriately specified *per pool* rather than *per plate*

### DIFF
--- a/labman/gui/handlers/process_handlers/pooling_process.py
+++ b/labman/gui/handlers/process_handlers/pooling_process.py
@@ -31,21 +31,7 @@ POOL_FUNCS = {
             'parameters': [('floor_vol', 'floor-vol-'),
                            ('floor_conc', 'floor-conc-'),
                            ('total_nmol', 'total-nm-'),
-                           ('size', 'lib-size-')]},
-    'floor': {'function': PoolingProcess.compute_shotgun_pooling_values_floor,
-              'parameters': [('floor_vol', 'floor-vol-'),
-                             ('floor_conc', 'floor-conc-'),
-                             ('total_nmol', 'total-nm-'),
-                             ('size', 'lib-size-')]},
-    # As everything, amplicon works differently here, we use this just for
-    # being able to retrieve the arguments
-    'amplicon': {'function': None,
-                 'parameters': [('dna_amount', 'dna-amount-'),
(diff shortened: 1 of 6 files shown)


labman/gui/handlers/process_handlers/pooling_process.py
```diff
@@ -166,67 +155,55 @@ class BasePoolHandler(BaseHandler):
     def _compute_pools(self, plate_info):
         plate_id = plate_info['plate-id']
         func_name = plate_info['pool-func']
+        plate_type = plate_info['plate-type']
         func_info = POOL_FUNCS[func_name]
         function = func_info['function']
+
+        print('foo')
```
ElDeveloper: :shipit:

labman/gui/handlers/process_handlers/pooling_process.py
```diff
@@ -302,23 +279,22 @@ def get(self):
...
             plate_names = plate_names.tolist()
 
-            if pool_func_data['function'] == 'amplicon':
-                pool_func_data['parameters']['epmotion-'] = process.robot.id
-                pool_func_data['parameters'][
-                    'dest-tube-'] = process.destination
         elif len(plate_ids) > 0:
             content_types = {type(Plate(pid).get_well(1, 1).composition)
                              for pid in plate_ids}
+            print(content_types)
```
ElDeveloper: :shipit:

PR #217, opened 2018-03-27
ENH: Improve grid space usage

### LINKED ISSUE #210: Cells with lots of text
If there is too much text in a cell, it just fits part of the name then adds “…”. This may vary from person to person, but I prefer to just make the cells wide enough so that I can see all the text without having to click on the individual cells. Is it possible to increase the cell width to allow for all/more characters to be seen?
And tiny side note, you can also see that the column headers (#1-12) are slightly cut off at the bottom. Can they be shifted up a bit?
screen shot 2018-02-22 at 09 42 29

ElDeveloper:
A handful of usability improvements in the main grid:

- Make the width of the columns adjustable.
- Improve the default width by 25% (from 80 to 100 pixels).
- Make the text in the column headers fully visible.

Animated gif for reference:

more-space

Fixes #210

### DIFF
--- a/labman/gui/static/css/labman.css
+++ b/labman/gui/static/css/labman.css
@@ -180,6 +180,11 @@ kbd.key {
     pointer-events: none;
 }
 
+/* make the header text be fully visible */
+.full-height-header {
+  height: 100% !important;
+}
+
 /*
  *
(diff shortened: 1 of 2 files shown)


wasade:
:+1:

PR #221, opened 2018-04-25
Prep work for issues 195 and 214: "notes" field on Process

AmandaBirmingham:
Added "notes" varchar(600) field to qiita.process table, updated process.py code to set/access, updated test_process.py to test. Also added a re-quantitation process (and associated concentrations) to test db.

### DIFF
--- a/labman/db/process.py
+++ b/labman/db/process.py
@@ -84,9 +84,10 @@ class Process(base.LabmanObject):
         return instance
 
     @classmethod
-    def _common_creation_steps(cls, user, process_date=None):
+    def _common_creation_steps(cls, user, process_date=None, notes=None):
         if process_date is None:
             process_date = date.today()
+
         with sql_connection.TRN as TRN:
             sql = """SELECT process_type_id
                      FROM qiita.process_type
@@ -95,10 +96,10 @@ class Process(base.LabmanObject):
             pt_id = TRN.execute_fetchlast()
 
             sql = """INSERT INTO qiita.process
-                        (process_type_id, run_date, run_personnel_id)
-                     VALUES (%s, %s, %s)
+                        (process_type_id, run_date, run_personnel_id, notes)
+                     VALUES (%s, %s, %s, %s)
                      RETURNING process_id"""
-            TRN.add(sql, [pt_id, process_date, user.id])
+            TRN.add(sql, [pt_id, process_date, user.id, notes])
             p_id = TRN.execute_fetchlast()
         return p_id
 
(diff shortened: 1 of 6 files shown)


AmandaBirmingham:
@tanaes Merging this pull request should give @ElDeveloper the tools he needs to let users store notes when they perform a quantification.  There is now an optional final "notes" parameter to QuantificationProcess.create (used in labman/gui/handlers/process_handlers/quantification_process.py QuantificationProcessHandler.post) and create_manual (used in labman/gui/handlers/process_handlers/pooling_process.py PoolPoolProcessHandler.post).

tanaes:
Brilliant, thank you! I'll take a look at this asap, maybe if @ElDeveloper can as well before merge?

labman/db/process.py
```diff
@@ -1751,22 +1773,44 @@ def create(cls, user, plate, concentrations):
...
                      VALUES (%s, %s, %s)"""
-            sql_args = []
-            layout = plate.layout
 
-            for p_row, c_row in zip(layout, concentrations):
-                for well, conc in zip(p_row, c_row):
-                    if well is not None:
-                        sql_args.append([well.composition.composition_id,
-                                         instance.id, conc])
+            if plate is not None:
```
tanaes: So this is basically uniting the plate (per composition) and pool (single composition of a pool) quantifications? And it infers what it's doing based on the presence of a value for `plate`?

AmandaBirmingham: @tanaes Yes, precisely.  The new _create method unites the previous contents of create and create_manual (the create and create_manual methods still exist, so as not to alter the public interface for the class, but they each just call _create under the hood with the appropriate set of inputs for their functionality).  The reason for uniting these two methods is that they duplicated most of their functionality (including several string-literal SQL queries, which is a maintenance and bug-finding nightmare!) and I would have had to modify each of them separately but in exactly the same way in order to allow storage of notes, which is a design red flag for me.

As you noted,  _create decide whether the quantification is per-plate or per-pool based on whether the plate argument is provided.  This is perhaps not ideal (because, for example, there's no validation happening to ensure that of the format of the quantifications argument matches what it should be given the presence or absence of the plate argument); however,  keep in mind that _create is *only* called from within create or create_manual.  Those two public methods are therefore responsible for making it clear to the user what the acceptable combination of inputs is for each of the two ways of doing a quantification, and _create is assuming it is being passed sensible inputs from one of those two gate-keepers.

tanaes: Awesome, thanks for finding that and for the clarification!

Issue #204, opened 2018-02-20
update sample sheet generation to handle project shortname


tanaes:
We're using the `sample_proj` field in the Sample Sheets as a unique reference for demultiplexing to assign demuxed fastq files automatically to their project folder.

We need to update the sample sheet generation to read in a per-sample value:
https://github.com/jdereus/labman/blob/0b36cfb63e8222f078f2479c9651e225864fb9c1/labman/db/process.py#L2577

And we may also need to add a column to the projects table to allow a project shortname in a format like 'PI_qiitaID' or similar.

AmandaBirmingham:
@tanaes Some clarification questions:

1) You note "we may also need to add a column to the projects table", but I can't find a projects table in the labman/qiita db!  Are we perchance talking about the study table? (I ask because I see that the fix for issue #32, Plates List screen (Project names), got the "project names" it displays from the qiita.study.title value). If not, has a project table been added at some point but didn't make it into the labman db patch scripts?  Or does it have a name other than projects?
2) If we *are* talking about the qiita.study table, is the project shortname already in there?  For example, is it the study_id field or the study_alias field?
3) Right now the sample_proj parameter to _format_sample_sheet_data is being set to the sequencing process's run name.  After we change sample_proj to instead hold the per-sample project shortname, does the sequencing process's run name need to be included anywhere (else) in the sample sheet output?

tanaes:
I don't think that this is a field that exists in Qiita yet necessarily. CC @jdereus — the thing is we need something that can link up to a consistent folder name on Barnacle. We've had some discussions about how best to do this but I don't know that we've formally adopted a consensus yet!

So, 1 & 2: Correct, it doesn't exist yet in the Qiita db; and 3: I think it would be good to add an additional column to the sample name titled "Project_Name" that includes the long name while the protected column "Sample_Project" contains the short name.

jdereus:
We need to be very careful on the definition of "consistent folder name" and how it applies to a sample sheet.  Reading some of this, it sounds like somehow we are expect to know the run folder information prior to being able to generate the sample sheet.

I am not aware of a project short name.  other than what we add in to the sample sheet.  This should be something along the lines of <PI>_<qiita_ID>.  That will allow for per project, per plate data aggregation when we do the bcl conversion.  Please correct me if I am misunderstanding the issue here.

JD

AmandaBirmingham:
[...]

PR #222, opened 2018-04-26
Fix amplicon pooling (issue #204)

tanaes:
This PR fixes issue #203, consolidating how amplicon and shotgun library plates are pooled. Changes in functionality:

- both shotgun and amplicon libraries can be pooled using either 'equal volume' or 'minimum volume' pooling
- for both library types, blanks can optionally be handled differently than samples, either by choosing to pool only a set number of blanks (automatically choosing the N highest-concentrated blanks), and/or by setting blank pooling values to a specific volume. Defaults are currently set differently for each library type, in line with current practices.
- estimates of the total pool concentration and volume are computed and displayed below the heatmap representation per pool
- test data now have more realistic numbers for concentration
- a number of small bug fixes in the interface


labman/gui/templates/library_pooling.html
```diff
@@ -169,35 +187,34 @@
...
+        } else {
+          $.each(poolParams[plateType][poolFunc], function(idx, paramElem) {
+            var $element = $('#' + paramElem['prefix'] + plateId);
+
+            var isEmpty = $element.val() === '' || $element.val() === null;
+            var isFieldDisabled = $element.prop('disabled')
+
+            disable_compute = ((isEmpty && !isFieldDisabled) || disable_compute)
+
+            //$('#compute-pool-btn').prop('disabled', (isEmpty && !isFieldDisabled) || isCompPoolBtnDisabled);
```
ElDeveloper: Can you remove this line?

tanaes: Done

tanaes:
@ElDeveloper If you're satisfied, can you merge? I want to start propagating these PRs into main.

tanaes:
actually @ElDeveloper I think we should merge #228 first, and then update this to be compatible.

ElDeveloper:
Should be ready for merge after tests pass.

tanaes:
Thanks @ElDeveloper! This looks good to me. Ok to merge?

ElDeveloper:
Good to go!

PR #228, opened 2018-05-03
Fix for Issue 214

### LINKED ISSUE #214: Plates > Normalize gDNA Plates arbitrarily selects which quantification to use
The Labman database supports storing multiple quantifications per well.  However, the existing code for normalizing gDNA plates does not allow the user to decide which of the existing quantifications for the plate to use; instead, it always picks the first one it retrieves from an unsorted SQL query:
screen shot 2018-03-14 at 11 50 43 am

In the above example, it picks quantification process 8, which happens to be the most recent one for the plate in question.  However, as far as I know, Postgres does not impose any default sorting on query results unless one explicitly uses an order by clause (https://stackoverflow.com/a/6586018) , so I don't think it is safe to assume that the most recent quantification will always be the one returned.  This could be very bad if the reason that multiple quantifications were done is that the first one was incorrect.

AmandaBirmingham:
Fixes #214 by adding support for Plate objects having quantitation_processes property rather than (incorrect) single quantitation_process, and user selecting the appropriate quantitation process for each plate during normalization and pooling. Also corrects spelling of ComputeLibraryPoolValuesHandler (was ComputeLibraryPoolValueslHandler).

### DIFF
--- a/labman/db/plate.py
+++ b/labman/db/plate.py
@@ -379,13 +379,14 @@ class Plate(base.LabmanObject):
             return process_module.Process.factory(TRN.execute_fetchlast())
 
     @property
-    def quantification_process(self):
-        """The quantification process(es) of the plate
+    def quantification_processes(self):
+        """The quantification process(es) applied to (wells on) the plate
 
         Returns
         -------
-        QuantificationProcess
-            The quantification process of the plate, if exists. None, otherwise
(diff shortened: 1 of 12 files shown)


ElDeveloper:
Looks great, thanks @AmandaBirmingham!

Issue #233, opened 2018-05-09
Divide by zero in _compute_amplicon_pool_values


AmandaBirmingham:
In db/process.py _compute_amplicon_pool_values, this line

https://github.com/jdereus/labman/blob/0dd5afef6a980ff25d6b420618828a2bddae383c/labman/db/process.py#L1890

sometimes generates:

RuntimeWarning: divide by zero encountered in true_divide

Is having a zero sample concentration valid?  If so, perhaps we should handle that condition rather than letting it spit warnings ...

tanaes:
Hmm, good point -- yes it's possible for these to be zero and we need to be
able to handle that.

tanaes:
...is it that line or is it here?

https://github.com/jdereus/labman/blob/0dd5afef6a980ff25d6b420618828a2bddae383c/labman/db/process.py#L2022-L2028

In PR #203 we actually should have removed any use of `_compute_amplicon_pool_values` method.

AmandaBirmingham:
No, the warning I was getting was definitely from the "return float(dna_amount) / sample_concs " line--but I saw it a while ago and just got around to logging it.  So maybe it is now gone :)

tanaes:
In any case it should be a problem and it helped me catch that deprecated
method, so thanks! :)

tanaes:
Fixed with #235

PR #235, opened 2018-05-09
handle warnings for divide by zero in pooling, fixes issue #233

tanaes:
This: 

1. temporary changes numpy behavior to 'ignore' divide-by-zero warnings in the `compute_pooling_values_minvol` method. This leaves calculated volumes as `inf`.

2. removes deprecated static method `_compute_amplicon_pool_values`

I think the default numpy divide behavior here is the desired behavior -- producing `inf` volume values for cases where the measured concentration is zero.

Currently, these values get replaced by the 'floor_vol' value when 'floor_conc' is > 0. Maybe we should also add a constraint that requires `floor_conc` to be > 0?

### DIFF
--- a/labman/db/process.py
+++ b/labman/db/process.py
@@ -1887,7 +1887,11 @@ class QuantificationProcess(Process):
         np.array of floats
             A 2D array of floats
         """
-        return float(dna_amount) / sample_concs
+        old_err = np.seterr(divide='ignore')
+        out = float(dna_amount) / sample_concs
+        np.seterr(**old_err)
+
+        return out
 
 
 class PoolingProcess(Process):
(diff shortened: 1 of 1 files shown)


labman/db/process.py
```diff
@@ -2013,6 +1995,9 @@ def compute_pooling_values_minvol(
         sample_vols: np.array of floats
             the volumes in nL per each sample pooled
         """
+
+        old_err = np.seterr(divide='ignore')
```
ElDeveloper: I would suggest using the context manager version of this, as it makes it clear what block of code this directive is for:

```python
with np.errstate(divide='ignore'):
    # divisions and more!
```

tanaes:
That's great, thanks!

ElDeveloper:
Thanks @tanaes

PR #238, opened 2018-05-10
Issue 204 fix

### LINKED ISSUE #204: update sample sheet generation to handle project shortname

AmandaBirmingham:
Fixes #204 by changing value of Sample_Project column in shotgun sample sheet to hold short projectname for each sample, or "Controls" for non-experimental samples (control, empty, blank, etc).


AmandaBirmingham:
ping for review :)

labman/db/tests/test_process.py
```diff
@@ -1573,6 +1575,13 @@ def test_generate_prep_information(self):
         exp = {Study(1): SHOTGUN_EXAMPLE}
         self.assertEqual(obs[Study(1)], exp[Study(1)])
 
+    # Ok, the Knight lab doesn't test private methods.  But I do :) AEB
```
ElDeveloper: Should this block be uncommented, as per the note? 😄

AmandaBirmingham: Urgh, I suppose I can't just summarily change the testing procedure, so I'll remove the private test.

PR #240, opened 2018-05-22
Issue 215: Change qiita.process from date to time-zone-aware timestamp

AmandaBirmingham:
As a knock-on, changed PrimerWorkingPlateCreationProcess.create to include hrs:minutes in the autogenerated external id for the plate, per Yoshiki's request. Also changed handling in that function of collision of external ids, bc old way could easily fail; now appends random number if default external id already exists.

Added "else" case to generate_sample_sheet so that if an unrecognized assay type is passed in, code doesn't silently fail to produce anything but instead raises an error.


labman/db/process.py
```diff
@@ -127,6 +132,37 @@ def _get_process_attr(self, attr):
...
+        #
+        # However, when (say) writing the datetime as a string, no such smart
+        # reasoning applies.  As long as you are running the system in a single
+        # physical location, this shouldn't be a problem--all your times will
+        # be as you expect.  But if you import times from ANOTHER location, or
+        # if heaven forbid you MOVE from one location to another that is in a
+        # different time zone, you could get string representations of dates
+        # that are very different than you expected.
+        #
+        # So, be very, very careful.
```
tanaes: 😱 I hope that this extremely helpful comment never needs to be used.

labman/db/process.py
```diff
@@ -382,14 +419,16 @@ def create(cls, user, primer_set, master_set_order, creation_date=None):
...
+                Process.get_date_format())
             primer_set_plates = primer_set.plates
             check_name = '%s %s' % (primer_set_plates[0].external_id,
                                     plate_name_suffix)
             if plate_module.Plate.external_id_exists(check_name):
                 # The likelihood of this happening in the real system is really
                 # low, but better be safe than sorry
-                plate_name_suffix = datetime.now().strftime('%Y-%m-%d %H:%M')
+                plate_name_suffix = "{0} {1}".format(plate_name_suffix,
+                                                     randrange(1000, 9999))
```
tanaes: Is this still necessary, given that we're already including the `%H:%M` in the suffix?

labman/db/process.py
ElDeveloper: 👍

labman/gui/test/test_plate.py
```diff
@@ -302,6 +304,17 @@ def test_get_plate_handler(self):
         self.assertEqual(response.code, 404)
 
         # Plate has multiple quantitation processes
+        # Note: cannot hard-code the date in the below known-good object
```
ElDeveloper: Great catch!!

PR #258, opened 2018-07-11
BUG: Fix broken copy-paste

### LINKED ISSUE #243: Copy and pasting in plates stopped working

ElDeveloper:
The assignment of the `this.blankNames` property was being done
asynchronously on a per-cell basis. This worked fine if the user inputed
a value by hand, as it gave enough time for the request to return.
However, when someone copy-pasted text into the grid, the request
wouldn't return in time and the `this.blankNames` property wouldn't be
correctly assigned leading to a ReferenceError.

This is now fixed by getting the contents of `this.blankNames` before
initializing the whole grid. This way every cell has access to the data
before instantiation.

Fixes #243

### DIFF
--- a/labman/gui/static/js/plateViewer.js
+++ b/labman/gui/static/js/plateViewer.js
@@ -131,9 +131,16 @@ PlateViewer.prototype.initialize = function (rows, cols) {
 
   // Fetch the blank names before initializing the grid. This way we only make
   // one request to get this data instead of one per cell instantiation.
-  var blanksRequest = $.ajax({
+  // When no search term is specified in the request, we get all available
+  // blank types.
+  var blanksRequest = $.get({
     dataType: 'json',
     url: '/sample/control?term=',
+    error: function(jqXHR, textStatus, errorThrown) {
+     bootstrapAlert('Could not fetch known <i>Control Types</i> needed for ' +
+                    'plating samples. Try reloading the page. <br>' +
+                    jqXHR.responseText, 'danger');
+    }
   });
 
   var sgCols = [];
--- a/labman/gui/static/js/plateViewer.js
+++ b/labman/gui/static/js/plateViewer.js
@@ -129,6 +129,13 @@ PlateViewer.prototype.initialize = function (rows, cols) {
                        cssClass: 'slick-header-column',
                        headerCssClass: 'full-height-header'}];
 
+  // Fetch the blank names before initializing the grid. This way we only make
+  // one request to get this data instead of one per cell instantiation.
+  var blanksRequest = $.ajax({
+    dataType: 'json',
+    url: '/sample/control?term=',
+  });
+
   var sgCols = [];
 
   for (var i = 0; i < this.cols; i++) {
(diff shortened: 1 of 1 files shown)


labman/gui/static/js/plateViewer.js
```diff
@@ -129,6 +129,13 @@ PlateViewer.prototype.initialize = function (rows, cols) {
                        cssClass: 'slick-header-column',
                        headerCssClass: 'full-height-header'}];
 
+  // Fetch the blank names before initializing the grid. This way we only make
+  // one request to get this data instead of one per cell instantiation.
+  var blanksRequest = $.ajax({
+    dataType: 'json',
+    url: '/sample/control?term=',
```
wasade: To make sure I understand this, by not requesting a term or set of terms, we get back all of the possible controls?

labman/gui/static/js/plateViewer.js
```diff
@@ -536,13 +544,9 @@ function SampleCellEditor(args) {
...
 
-  this.blanknames = [];
-  $.ajax({
-    url: '/sample/control?term=',
-    success: function (result) {
-      that.blankNames = result;
-    }
-  });
+  // Get the parsed values from the request. Alternatively, if the request
+  // didn't complete yet use an empty list as a fallback.
```
wasade: Is it a failure if the request does not complete?

labman/gui/static/js/plateViewer.js
wasade: Okay, thanks. Would it be possible to add a comment indicating as such as the behavior is implicit?

labman/gui/static/js/plateViewer.js
wasade: I think so, yes. Is it possible to block until the ajax completes followed by a test of a status code?

ElDeveloper: Blocking requests are frowned upon and deprecated (or on their way out). This [link](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Synchronous_and_Asynchronous_Requests) has information (see under Synchronous Requests).

For now, I'll add some error handling and let the user know if they need to reload the page to try again.

labman/gui/static/js/plateViewer.js
ElDeveloper: Will do.

labman/gui/static/js/plateViewer.js
ElDeveloper: I've added some error handling (a message at the top of the screen describing the problem and a possible workaround). Note that for this test I modified the URI so the request pointed to a non-existant route hence the 404.

screen shot 2018-07-25 at 12 50 27 pm

wasade: :+1:

wasade:
I'm :+1: pending travis. @AmandaBirmingham, would you like a second review or good if green w/ travis?

AmandaBirmingham:
@wasade I looked over the changes and the discussion, and I don't have any useful comments.  If travis green-lights this, so do I :)

ElDeveloper:
Thanks!

PR #264, opened 2018-07-25
Fixes issue 251: Extend pool list to offer re-download of pool files for plate pools

AmandaBirmingham:
Pool list now looks like the below, with pool file download buttons for plate pools and none for sequencing pools:

screen shot 2018-07-25 at 7 03 37 am

### DIFF
--- a/labman/db/composition.py
+++ b/labman/db/composition.py
@@ -1022,10 +1022,14 @@ class PoolComposition(Composition):
             [{'pool_id': int, 'external_id': string}]
         """
         with sql_connection.TRN as TRN:
-            sql = """SELECT pool_composition_id, external_id
+            sql = """SELECT pool_composition_id, pooling_process_id, 
+                     external_id 
                      FROM qiita.pool_composition
                         JOIN qiita.composition USING (composition_id)
                         JOIN qiita.tube USING (container_id)
+                        JOIN qiita.process ON 
+                          upstream_process_id = qiita.process.process_id
+                        JOIN qiita.pooling_process USING (process_id) 
                      ORDER BY pool_composition_id"""
             TRN.add(sql)
             return [dict(r) for r in TRN.execute_fetchindex()]
--- a/labman/db/composition.py
+++ b/labman/db/composition.py
@@ -1029,6 +1029,13 @@ class PoolComposition(Composition):
 
     @staticmethod
     def get_pools():
+        """Return list of all pool compositions as PoolComposition objects
+
+        Returns
+        -------
+        list of labman.db.composition.PoolComposition
+            Ordered by pool_composition_id
+        """
         with sql_connection.TRN as TRN:
             sql = """SELECT pool_composition_id
                      FROM qiita.pool_composition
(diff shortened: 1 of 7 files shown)


labman/db/composition.py
```diff
@@ -1012,23 +1012,32 @@ class PoolComposition(Composition):
...
-            The list of pool information with the structure:
-            [{'pool_id': int, 'external_id': string}]
+        Parameters
+        ----------
+        components: list-like of Compositions
         """
+
+        # This logic is taken from the original
+        # process.PoolingProcess.generate_pool_file method
+        # TODO: Someday: I don't like how this checks only the *first* of the
```
antgonza: Should the TODO be issues in the repo and add the issue number to this comment for easier tracking? This questions applies to all the TODOs in this PR.

labman/db/composition.py
```diff
@@ -1012,23 +1012,32 @@ class PoolComposition(Composition):
...
+        # components of the composition!
+        comp = components[0]
+        return comp.__class__
+
+    @staticmethod
+    def get_pools():
```
antgonza: docstring?

labman/db/composition.py
```diff
@@ -1087,6 +1102,13 @@ def raw_concentration(self):
             res = TRN.execute_fetchindex()
             return res[0][0] if res else None
 
+    @property
+    def is_plate_pool(self):
```
antgonza: docstring?

labman/db/process.py
```diff
@@ -2378,17 +2380,18 @@ def generate_pool_file(self):
         str
             The contents of the pool file
         """
-        comp = self.components[0][0]
-        if isinstance(comp, composition_module.LibraryPrep16SComposition):
+        component_compositions = [x[0] for x in self.components]
```
antgonza: I think the proper way to test this is via of isinstance ....

labman/gui/test/test_pool.py
```diff
@@ -0,0 +1,28 @@
...
+class TestPoolHandlers(TestHandlerBase):
+    def test_get_plate_list_handler(self):
+        response = self.get('/pool_list')
+        self.assertEqual(response.code, 200)
+        obs = json_decode(response.body)
+        self.assertCountEqual(obs.keys(), ['data'])
+        obs_data = obs['data']
+        self.assertEqual(len(obs_data), 3)
+        self.assertEqual(obs_data[0], [1, 'Test Pool from Plate 1', True, 1])
+        self.assertEqual(obs_data[1], [2, 'Test sequencing pool 1', False, 2])
```
antgonza: Just out of curiosity, why not test obs_data[2]?

labman/db/process.py
AmandaBirmingham: In this case, after the refactor, the code is actually comparing classes rather than comparing an instance to a class; in the former case, isinstance doesn't work so I followed the guidance at https://stackoverflow.com/a/23012248 re comparing classes directly.  I suppose one might argue that the approach of comparing classes is undesirable ...

labman/gui/test/test_pool.py
AmandaBirmingham: Just lazily following the approach of the code I copied from, which just tested the overall number of items and only a few of the individuals that were known to be interesting.  However, since there are only three items here, I went ahead and added the third test :)

antgonza:
After offline review, we decided that it was good to merge and that @AmandaBirmingham will add the links to the issues/TODOs in a future PR.

Issue #239, opened 2018-05-10
Include lane column/number in sample sheets for multilane sequencers


tanaes:
Original issue name: Need per-lane pool assignment in 'Prepare sequencing run' interface

Currently, there is no mechanism to assign sequence pools to individual lanes (e.g. 8 lanes of HiSeq4000 or 2 lanes of HiSeq2500). 

I'd imagine that we'd want to have the interface load the number of lane slots appropriate to the sequencing instrument once that instrument is loaded, and then allow just a single pool to be added per lane.

AmandaBirmingham:
It appears this issue is a bit different than originally thought.  The existing code in labman.db.process.SequencingProcess.create expects to take in "The pools being sequenced, in lane order", and it does store the lane numbers associated with each pool (as implied by that order) into the qiita.sequencing_process_lanes table.  I checked with Greg H. today, and he says that he is ok (at least for now) with the lane number for each pool being implied by the order in which it is added to the sequencing run.

Unfortunately, the existing code DOESN'T output the lane information into the sample sheets produced, which is a bug.  Greg indicates that for both 16s and shotgun sample sheets, for sequencers that have more than one lane, each row should start with a "Lane" column holding the (1-based) lane number.  For 16s (amplicon) sample sheets (only), there should be one row per lane and each row should have the prescribed string of Ns in the index column.

AmandaBirmingham:
Did a bit of investigation here; the labman.db.process.SequencingProcess. _generate_shotgun_sample_sheet method IS lane-aware (via _format_sample_sheet_data).  However, _generate_amplicon_sample_sheet is NOT, and will need a bit of a rewrite to make it so.

AmandaBirmingham:
@qiyunzhu please take a look at the details above watch this space in case I get any more info from Greg :)

qiyunzhu:
@AmandaBirmingham Yes I am tracking this thread :) Thanks for all the information!

qiyunzhu:
Hi @tanaes @AmandaBirmingham I created a pull request #268 to address this issue.

Currently, the "Lane" column has been successfully added to the amplicon sample sheet. This was achieved by modifying the `_generate_amplicon_sample_sheet` function.

However I felt that this function was originally flawed. It only generates one row of the whole sample sheet, rather than the entire sample list, as the counterpart `_generate_shotgun_sample_sheet` does.

Did I miss something? I would appreciate if either of you can confirm!

AmandaBirmingham:
[...]

PR #268, opened 2018-07-25
added column "Lane" to amplicon sample sheet

qiyunzhu:
To address issue #239 . @AmandaBirmingham @tanaes 

Currently, the "Lane" column has been successfully added to the amplicon sample sheet. This was achieved by modifying the `_generate_amplicon_sample_sheet` function.

However I felt that this function was originally flawed. It only generates one row of the whole sample sheet, rather than the entire sample list, as the counterpart `_generate_shotgun_sample_sheet` does.


qiyunzhu:
Hi @AmandaBirmingham I think I have tackled all the requirements. Thanks in advance for reviewing! A separate email with example result files will be sent out.

labman/db/process.py
```diff
@@ -2861,7 +2872,8 @@ def _generate_shotgun_sample_sheet(self):
                 bcl2fastq_sample_ids, i7_names, i7_sequences, i5_names,
                 i5_sequences, sample_proj_values, wells=wells,
                 sample_plates=sample_plates, description=samples_contents,
-                lanes=[lane], sep=',', include_header=include_header))
+                lanes=[lane], sep=',', include_header=include_header,
+                include_lane=(self.lane_count > 1)))
```
AmandaBirmingham: Currently, the "self.lane_count > 1" check occurs three different places, in two different functions. I am worried this leaves us open to future bugs due to someone only updating some of the occurrences.

I suggest you turn the lane_count int property into an include_lane boolean property (after all, the only time the lane_count is used is to determine whether to include the header.)  Then you can just reference that property rather than hardcoding the check into multiple locations.

labman/db/process.py
```diff
@@ -2948,12 +2960,15 @@ def _generate_amplicon_sample_sheet(self):
...
             The illumina-formatted sample sheet
         """
-        fixed_run_name = SequencingProcess._bcl_scrub_name(self.run_name)
-        data = (
-            'Sample_ID,Sample_Name,Sample_Plate,Sample_Well,I7_Index_ID,'
-            'index,Sample_Project,Description,,\n'
-            '%s,,,,,NNNNNNNNNNNN,,,,,' % fixed_run_name)
-        return self._format_sample_sheet(data)
+        data = ['%sSample_ID,Sample_Name,Sample_Plate,Sample_Well,'
+                'I7_Index_ID,index,Sample_Project,Well_Description,,'
```
AmandaBirmingham: Could you add a comment on why the Well_Description (rather than just Description) column name was requested, so, institutionally, we don't forget?

labman/db/tests/test_process.py
```diff
@@ -1477,7 +1497,7 @@ def test_format_sample_sheet(self):
         self.assertEqual(exp_sample_sheet, obs_sample_sheet)
 
     def test_generate_sample_sheet(self):
-        # Sequencing run
+        # Amplicon run
         tester = SequencingProcess(1)
```
AmandaBirmingham: I do not see any tests that AMPLICON sample sheet generation (which uses _generate_amplicon_sample_sheet rather than _format_sample_sheet_data by way of _generate_shotgun_sample_sheet) works correctly in the multi-lane case as well as the single-lane case.  Could we please add one?

qiyunzhu:
@AmandaBirmingham Thanks for reviewing! I will take care asap.

qiyunzhu:
Hi @AmandaBirmingham I addressed your first two comments. But for the third one I have difficulties. I can't figure out how to hard code a `SequencingProcess` object, as this [line](https://github.com/jdereus/labman/blob/master/labman/db/tests/test_process.py#L1212) does. Do you have some ideas?

AmandaBirmingham:
@qiyunzhu Uggh.  We shall have to modify populate_test_db.sql.  I suspect it will go fastest if we do it together; email me to let me know when you're available today to pair-program it.

qiyunzhu:
Hi @AmandaBirmingham I figured out a workaround. Everything is okay now!

labman/db/process.py
```diff
@@ -2948,12 +2961,17 @@ def _generate_amplicon_sample_sheet(self):
...
-            '%s,,,,,NNNNNNNNNNNN,,,,,' % fixed_run_name)
-        return self._format_sample_sheet(data)
+        # the "Description" => "Well_Description" change was for the
+        # compatibility with EBI submission
+        data = ['%sSample_ID,Sample_Name,Sample_Plate,Sample_Well,'
+                'I7_Index_ID,index,Sample_Project,Well_Description,,'
+                % ('Lane,' if self.include_lane else '')]
+        for pool, lane in self.pools:
+            data.append('%s%s,,,,,NNNNNNNNNNNN,,%s,,,'
+                        % (('%s,' % lane if self.include_lane else ''),
```
wasade: Would it be possible to enclose `lane if self.include_lane else ''` in parentheses? The execution here is entirely dependent on the order of operations, and it's not clear on a static review if the evaluation is first `'%s,' % lane` which would make that "expr1" or if the evaluation is first `lane if self.include_lane else ''`.

labman/db/tests/tester.py
```diff
@@ -416,8 +416,8 @@ def stress_tests(num_plates):
 ReverseComplement,0
 
 \[Data\]
-Sample_ID,Sample_Name,Sample_Plate,Sample_Well,I7_Index_ID,index,Sample_Project,Description,,
-New_sequencing_run_\d{4}-\d{2}-\d{2}_\d{2}_\d{2}_\d{2}_\d{6},,,,,NNNNNNNNNNNN,,,,,"""  # noqa
+Sample_ID,Sample_Name,Sample_Plate,Sample_Well,I7_Index_ID,index,Sample_Project,Well_Description,,
+New_pool_name_\d{4}-\d{2}-\d{2}_\d{2}_\d{2}_\d{2}_\d{6},,,,,NNNNNNNNNNNN,,\d+,,,"""  # noqa
```
wasade: What's the `\d+` for?

labman/gui/templates/sequencing.html
```diff
@@ -64,7 +64,7 @@
     $buttonElem.append("<span class='glyphicon glyphicon-remove'></span>")
     $aElem.append($buttonElem);
     $('#pool-list').append($aElem);
-    $('#addBtnPool' + poolId).prop('disabled', true);
+    // $('#addBtnPool' + poolId).prop('disabled', true);
```
wasade: Can this comment and the one on line 76 be removed please?

labman/db/tests/tester.py
qiyunzhu: `\d+` is the `composition_id` property of a pool, which is a number (e.g., `3079`). During yesterday's discussion we decided to let "Well_Description" be the identifier of the pool.

qiyunzhu:
Hi @wasade Thanks for reviewing! I have made edits and comments.

labman/db/process.py
qiyunzhu: Good suggestion!

labman/gui/templates/sequencing.html
qiyunzhu: Done.

labman/db/process.py
```diff
@@ -2968,7 +2968,7 @@ def _generate_amplicon_sample_sheet(self):
                 % ('Lane,' if self.include_lane else '')]
         for pool, lane in self.pools:
             data.append('%s%s,,,,,NNNNNNNNNNNN,,%s,,,'
-                        % (('%s,' % (lane if self.include_lane else '')),
+                        % (('%s,' % lane) if self.include_lane else '',
                            self._bcl_scrub_name(pool.container.external_id),
```
AmandaBirmingham: I do not understand this change.  @wasade requested "enclose lane if self.include_lane else '' in parentheses", which is not the change that was made here.  In fact, now the ternary conditional (lane if self.include_lane else '') is broken up by a parenthesis, which just doesn't seem right.  @wasade , can you weigh in?

qiyunzhu: Hi @AmandaBirmingham , the edit Daniel suggested will produce a different result. But his rationale was right, that my original version was visually confusing. Therefore I edited this part to make it visually clean.

qiyunzhu: The correct logic is:
```
if self.include_lane:
    return '%s,' % lane
else:
    return ''
```

AmandaBirmingham:
Ahh!  Now I gotcha!  Thank you for clarifying for me :)

On Jul 27, 2018, at 12:21 PM, Qiyun Zhu <notifications@github.com<mailto:notifications@github.com>> wrote:


@qiyunzhu commented on this pull request.

________________________________

In labman/db/process.py<https://github.com/jdereus/labman/pull/268#discussion_r205875461>:

> @@ -2968,7 +2968,7 @@ def _generate_amplicon_sample_sheet(self):
                 % ('Lane,' if self.include_lane else '')]
         for pool, lane in self.pools:
             data.append('%s%s,,,,,NNNNNNNNNNNN,,%s,,,'
-                        % (('%s,' % (lane if self.include_lane else '')),
+                        % (('%s,' % lane) if self.include_lane else '',
                            self._bcl_scrub_name(pool.container.external_id),


The correct logic is:

if self.include_lane:
    return '%s,' % lane
else:
    return ''

labman/db/process.py
AmandaBirmingham: Ahh!  Now I gotcha!  Thank you for clarifying for me :)

PR #269, opened 2018-07-27
Adds sample plating notes to the UI

ElDeveloper:
The user can now add notes to a plate during the plating process. These notes are later viewable when selecting plates in the extract gDNA UI. 

Here's an example of the new functionality. First I add a study to the plate, then add a sample, then when the process has been created the notes box is added to the UI. The user can add any text and save it with the "save" button.

Afterwards in the extract gDNA UI when a plate is selected the notes are shown at the bottom. See this animated gif for an example:

notes

### DIFF
--- a/labman/db/process.py
+++ b/labman/db/process.py
@@ -225,6 +225,11 @@ class _Process(Process):
     def notes(self):
         return self._get_attr('notes')
 
+    @notes.setter
+    def notes(self, value):
+        """Updates the notes of the process"""
+        self._set_attr('notes', value)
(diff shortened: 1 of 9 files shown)


ElDeveloper:
@tanaes @AmandaBirmingham, if this looks good to you in terms of the functionality (see animated gif above) would anyone of you mind merging?

AmandaBirmingham:
Done!

ElDeveloper:
Thanks so much! 🎉

PR #280, opened 2018-07-31
fix #262

### LINKED ISSUE #262: Populate real labman db with needed real equipment/etc records
Greg H. pointed out that the list of sequencers available on the Prepare Sequencing Run page in the labman test database is not complete and will need to be added to before soft launch.  There may be other such set-up that needs to be done (see the "Database Preparation" section of labman/db/support_files/labman_DB_schema_description.pdf for a list of the sorts of information that need to be pre-populated before use).

antgonza:
ended up adding the google spreed sheet query cause I had most of it done ... and it's easier to remove than reimplement

### DIFF
--- a/scripts/load-values
+++ b/scripts/load-values
@@ -1,0 +1,94 @@
+#!/usr/bin/env python
+
+# ----------------------------------------------------------------------------
+# Copyright (c) 2017-, labman development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# ----------------------------------------------------------------------------
+
+import click
+
+
+@click.group()
+def load_values():
+    pass
+
+
+@load_values.command()
+def load_values():
+    """"""
+    import pandas as pd
+    import requests
+    from io import BytesIO
+    from labman.db.sql_connection import TRN
+
+    # retriving spreadsheet
+    url = ('https://docs.google.com/spreadsheet/ccc?key=1knFJ6YC2gSNZBZhjJ_k6k'
+           'qTiQR8WQv3_HAK7vmmBk_w&output=csv')
+    request = requests.get(url)
+    if request.status_code != 200:
+        raise ValueError(
+            'There was an error with the request: %s, error %s' % (
+                url, request.reason))
+
+    # converting spreadsheet info to a pandas dataframe
+    df = pd.read_csv(BytesIO(request.content))
+
+    # unique machines
+    machines = df.valid_types.dropna().unique().tolist()
+
+    # unique tools with type_is_valid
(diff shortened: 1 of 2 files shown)


antgonza:
Thanks for review! It will be much faster turn around if we have a detailed list of changes vs. a readme (which to be honest haven't read! 😞).

Now, in reality, we will never use a script to recover a gold standard database snapshot in production as we will normally have snapshots of that state and can push to production, if needed ... hopefully never. Adding @jdereus, in case he has something to add.

josenavas:
Just my 0.02

The Database Preparation steps that Amanda is referring to should be included in the db_patch_manual.sql. Note that the goal of that file is to accomplish exactly what Amanda is requesting, an SQL file that allows us to load a clean DB from scratch. The values that were added the ones given by Rodolfo at that time.

Also, agree with Antonio's comment: once the DB is loaded, we should not need to wipe-out all the data and start over, but rather load previous backups.

To add additional equipment I think there is already a GUI to do that (so need to maintain external docs). Same for users, once 1 user has been added he can manage the rest through the GUI.

For the equipment type and the other records that Amanda are pointing out, they should only be changed when changes on the worlflow and/or DB structure happen, so they should be changed using the classical patch mechanism that has been used in Qiita and other systems.

Hope this helps!

AmandaBirmingham:
I am confident that you and @tanaes can generate the needed change list from the numbered table population steps in the document; they are on pages 2-4.  Some of them may be handled by other scripts already in labman (e.g., db_patch_manual.sql) but that needs to be checked for each of them and if filling is already being done in an existing script, it will be necessary to check that they are being filled with exactly what the wet lab expects.

tanaes:
I think the issue here is that this script really was requested to fill a sort of intermediate need, for initial spin-up. The values in the included sql files are really representative values -- one of each type of thing, so that tests can be run. And as Jose says, there's an interface available for piecemeal addition of new equipment, which should be used for example when the wet lab gets a new pipetter head for the EP Motion robots. 

IMHO I think this is something that will be useful primarily during the initial testing phase, and be replaced by web interface + restoring from backups once things are live. Does that make sense?

scripts/load-values
```diff
@@ -0,0 +1,94 @@
...
+                 WHERE NOT EXISTS (
+                    SELECT 1
+                    FROM labman.equipment
+                    WHERE external_id = %s )"""
+        for idx, row in tools.iterrows():
+            _id = equipment_type[row['type']]
+            name = row['tool_name_ID']
+            TRN.add(sql, [name, _id, name])
+
+        # finally insert all admins to labman.labmanager_access
```
tanaes: Does this really want to be part of this script? Or should the script be limited to just equipment loading?

Issue #272, opened 2018-07-27
Duplicate key error when pasting platemap into sample plating page


AmandaBirmingham:
NB: So far only observable against qiita-rc database, not against labman test database.

` Error running SQL query: Query: INSERT INTO qiita.plate (external_id, plate_configuration_id) VALUES (%s, %s) RETURNING plate_id Arguments: ['Manary_9', '2'] Error: duplicate key value violates unique constraint "idx_plate" DETAIL: Key (external_id)=(Manary_9) already exists.`

Full error trace here:
[manary_9_plate_error[3].pdf](https://github.com/jdereus/labman/files/2237440/manary_9_plate_error.3.pdf)

antgonza:
Sorry for multiple messages, note that this error is shows a few times in the logs but the specific one based on the example in the original opening of the issue is:
```
[E 180727 14:19:54 web:1670] Uncaught exception POST /process/sample_plating (137.110.53.215)
    HTTPServerRequest(protocol='https', host='qiita-rc.ucsd.edu:5573', method='POST', uri='/process/sample_plating', version='HTTP/1.1', remote_ip='137.110.53.215')
    Traceback (most recent call last):
      File "/home/qiita_test/labman/labman/db/sql_connection.py", line 586, in _execute
        cur.execute(sql, sql_args)
      File "/home/qiita_test/miniconda3/envs/labman/lib/python3.5/site-packages/psycopg2/extras.py", line 144, in execute
        return super(DictCursor, self).execute(query, vars)
    psycopg2.IntegrityError: duplicate key value violates unique constraint "idx_plate"
    DETAIL:  Key (external_id)=(Manary_9) already exists.
    
    
    During handling of the above exception, another exception occurred:
    
    Traceback (most recent call last):
      File "/home/qiita_test/miniconda3/envs/labman/lib/python3.5/site-packages/tornado/web.py", line 1590, in _execute
        result = method(*self.path_args, **self.path_kwargs)
      File "/home/qiita_test/miniconda3/envs/labman/lib/python3.5/site-packages/tornado/web.py", line 3006, in wrapper
        return method(self, *args, **kwargs)
      File "/home/qiita_test/labman/labman/gui/handlers/process_handlers/sample_plating_process.py", line 24, in post
        user, PlateConfiguration(plate_config_id), plate_ext_id)
      File "/home/qiita_test/labman/labman/db/process.py", line 263, in create
        plate = plate_module.Plate.create(plate_ext_id, plate_config)
[...]

PR #281, opened 2018-07-31
Sample plating improvements

ElDeveloper:
This PR was written by @AmandaBirmingham and me. Briefly:

- This PR addresses #272: The error was caused because the plate UI incorrectly allowed for the addition of multiple samples before the plate was created. The UI has now been modified to make the plate creation step a requirement before the grid is displayed.

- This also fixes a problem where listing plates without study identifiers would result in the page crashing or no plates being loaded.

- Fixes a problem with viewing an existing plate where 384 plates would now be fully displayed on screen.

- Fixes a problem where the notes would note be loaded when viewing an existing plate.

- This adds a placeholder for the *Notes* widget in the sample plating interface.

Animated GIF:

copy-pasting

### DIFF
--- a/labman/gui/static/js/plate.js
+++ b/labman/gui/static/js/plate.js
@@ -116,5 +116,38 @@ function change_plate_configuration() {
     // reset the container before updating the grid configuration
     $('#plate-map-div').empty().height(0);
     pv = new PlateViewer('plate-map-div', undefined, undefined, $opt.attr('pm-data-rows'), $opt.attr('pm-data-cols'));
+    // Enable the plate create button now that a plate config is selected
+    $('#createPlateBtn').prop('disabled', false);
   }
+}
+
+function createPlate(){
+      var plateName = $('#newNameInput').val().trim();
+      var plateConf = $('#plate-conf-select option:selected').val();
+      $.post('/process/sample_plating', {'plate_name': plateName, 'plate_configuration': plateConf}, function (data) {
+        var plateId = data['plate_id'];
+        var processId = data['process_id'];
+
+        $('#plateName').prop('pm-data-plate-id', plateId);
+        $('#plateName').prop('pm-data-process-id', processId);
+        // Once the plate has been created, we can disable the plate config select
+        $('#plate-conf-select').prop('disabled', true);
+
+        // reset the container before updating the grid configuration
+        $('#plate-map-div').empty().height(0);
+        var $opt = $('#plate-conf-select option:selected');
+        var pv = new PlateViewer('plate-map-div', undefined, undefined, $opt.attr('pm-data-rows'), $opt.attr('pm-data-cols'));
+        pv.plateId = plateId;
+        pv.processId = processId;
+        // we can only instantiate the notes box when we have a process id
+        pv.notes = new NotesBox(pv.container.parent(),
+                                  '/process/sample_plating/notes',
+                                  processId);
+
+        // Disable the plate create button
+        $('#createPlateBtn').prop('disabled', true);
+
+        // Show the plate details div
+        $('#plateDetails').prop('hidden', false);
+      });
 }
--- a/labman/gui/static/js/plate.js
+++ b/labman/gui/static/js/plate.js
@@ -121,33 +121,47 @@ function change_plate_configuration() {
   }
 }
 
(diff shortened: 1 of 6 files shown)


labman/gui/static/js/plate.js
```diff
@@ -116,5 +116,52 @@ function change_plate_configuration() {
...
+    $('#plateName').prop('pm-data-plate-id', plateId);
+    $('#plateName').prop('pm-data-process-id', processId);
+    // Once the plate has been created, we can disable the plate config select
+    $('#plate-conf-select').prop('disabled', true);
+
+    // reset the container before updating the grid configuration
+    $('#plate-map-div').empty().height(0);
+    var $opt = $('#plate-conf-select option:selected');
+    var pv = new PlateViewer('plate-map-div', undefined, undefined, $opt.attr('pm-data-rows'), $opt.attr('pm-data-cols'));
+    pv.plateId = plateId;
```
AmandaBirmingham: I realized that there had better be a comment on these lines, or the next time I see them, I will try to "fix" them the way I did yesterday (put the plateId and the processId into the two undefined values in the "new PlateViewer" call), which I currently remember doesn't work.  Could you add this comment above these lines?

// Create a new PlateViewer object *without* passing in plateId and processId,
// because if those are passed in to the object initializer, code assumes the  
// plate has already been filled--and when it can't find contents, it populates
// all the wells with blanks :(.  Setting the plateId and processId after object
// initialization sets the necessary state but avoids this behavior.

ElDeveloper: 👍 should be added now.

AmandaBirmingham:
Ping @wasade and @antgonza : this PR is on hold waiting for code review

labman/gui/templates/plate_list.html
```diff
@@ -101,7 +101,11 @@
           var deleteButton = '<a onclick="discardPlate(' + row[0] + ', this)" class="btn btn-danger btn-circle-small">' +
                                '<span class="glyphicon glyphicon-remove" data-toggle="tooltip" title="Remove plate"></span>' +
                              '</a> ';
-          newData.push([chBox, row[0], row[1], row[2].join('<br/>'), deleteButton]);
+          var studiesList = '';
+          if (row[2] !== null) {
```
wasade: Would it be possible to have a comment noting what index 2 is?

labman/gui/templates/plate_list.html
```diff
@@ -101,7 +101,11 @@
...
+              studiesList = row[2].join('<br />');
+          }
+          newData.push([chBox, row[0], row[1], studiesList, deleteButton]);
```
wasade: Same same

AmandaBirmingham:
Ping @wasade please take a look at the requested changes and approve or not :)

labman/gui/templates/plate.html
```diff
@@ -115,37 +125,42 @@
...
-<div id='plate-map-div' class='spreadsheet-container'></div>
+    <!-- Plate map div -->
+    <h4>Plate map</h4>
 
-<h4>Well comments</h4>
-<div id='well-plate-comments' class="comment-box"></div>
+    <!-- Controls div -->
+    <div><h5>Controls description</h5>
+      <p>Press <kbd class='key'>ESC</kbd> to enter cell <i>selection mode</i></p>
+      <p>Press <kbd class='key'>ENTER</kbd> or double click a cell to enter <i>edit mode</i></p>
```
tanaes: Worth adding the instruction to right-click on a well to add comments to that well? Otherwise hard to discover.

labman/gui/static/js/plateViewer.js
```diff
@@ -263,33 +262,8 @@ PlateViewer.prototype.initialize = function (rows, cols) {
...
-        });
-    } else {
-      // The plate already exists, simply plate the sample
-      that.modifyWell(row, col, content);
-    }
+    // The plate already exists, simply plate the sample
+    that.modifyWell(row, col, content);
   });
 
   // When the user right-clicks on a cell
```
tanaes: Is it possible to make the well comment text field automatically scoped for entering text when it's displayed? As it stands now it requires an additional mouse click to activate.

PR #317, opened 2018-08-09
 fixes #291 : Fix parsing of quantitation files

### LINKED ISSUE #291: Fix parsing of quantitation files
Handle type of newlines in file created by wetlab (windows, maybe?)
 - Without this, receive an uncaught exception when try to upload quantitation file
Automatically replace "nan" concentrations with 0 when parsing quantitation files
 - Without this, nan wells display as white, AND received an error when clicking Confirm

Interpret any values listed as "<0.00" as 0.00
Interpret any values listed as ">(some numeric value)" as "(some numeric value)"
Interpret any values listed as a bunch of question marks (when you overflow the sensor; usually sample being too concentrated) as the highest quantification value in that particular file.
 - For all three of these, the wet lab currently does this manually

Raised in shotgun walkthrough 07/31/2018

AmandaBirmingham:
Prevent parsing exception caused by CR newlines
Interpret any values listed as "<0.00" as 0.00
Interpret any values listed as ">some numeric value" as some numeric value
Interpret any concentrations listed as a bunch of question marks (when you overflow the sensor; usually sample being too concentrated) as the highest concentration value in that particular file
Raise an error if, after these fixes, there are any concentration values that are still NaN

### DIFF
--- a/labman/db/process.py
+++ b/labman/db/process.py
@@ -1702,8 +1702,8 @@ class QuantificationProcess(Process):
 
         Parameters
         ----------
-        contents: fp or open filehandle
-            pico quant file
+        contents : str
+            The contents of the pico green plate reader output
         sep: str
             sep char used in quant file
         conc_col_name: str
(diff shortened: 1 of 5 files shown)


labman/db/process.py
```diff
@@ -1714,18 +1714,60 @@ def _parse_pico_csv(contents, sep='\t',
...
 
+        # find the highest concentration in the file and replace all overflow
+        # concentrations with that value
+        max_concentration = pico_df[conc_col_name].max()
+        pico_df.loc[overflow_mask, conc_col_name] = max_concentration
+
+        # if there are any NaN concentrations left, there's a problem with the
+        # parsing, so throw an error
+        if pico_df[conc_col_name].isnull().any():
+            raise ValueError("Some concentrations in pico green quantitation "
```
antgonza: Is this actually reported to user in the GUI? In other words, when you do this via the GUI and leave NaNs do you see this error?

----- As this is not part of this PR, if the answer is no; let's just open an issue.

antgonza: Thanks. Let's open an issue but make it the least priority possible.

AmandaBirmingham: Will do ;)

Issue #296, opened 2018-08-01
Update amplicon sample sheet with currently used values


AmandaBirmingham:
In [Header]:
Change value associated with "Assay" from "Amplicon" to "TruSeq HT"
Change value associated with "Chemistry" from "Default" to "Amplicon"
In [Settings]:
Add "Adapter" field and appropriate value as DNA
Add "AdapterRead2" field and appropriate value as DNA
(Scope of this issue includes FINDING OUT from wet lab what the appropriate values are and figuring out where to get the in Labman db).
In [Data]:
Check with Greg whether I5_Index_ID and index2 fields should be added to output (after index column)
Note: do NOT change "Well_Description" back to "Description" : Gail Ackermann requested this change as part of PR 268. 

Raised in amplicon walkthrough 07/30/2018

AmandaBirmingham:
Per Greg Humphrey: "The adapters are always the same and are located on our primers" and "the adapters are the same for 16S, 18S, and ITS".  Thus, the values for Adapter and Adapter2 will be the same in every amplicon sample sheet, regardless of primer set, etc.

qiyunzhu:
Hi @AmandaBirmingham I have tackled this issue. See PR #318 . I talked with Greg and understood their requirement in the sample sheet format.

PR #318, opened 2018-08-09
modifed amplicon sample sheet

qiyunzhu:
Addressed issue #296 .


labman/db/process.py
```diff
@@ -2807,8 +2814,10 @@ def _format_sample_sheet(self, data, sep=','):
             'Workflow{sep}{Workflow}\nApplication{sep}{Application}\n'
             'Assay{sep}{Assay}\nDescription{sep}{Description}\n'
             'Chemistry{sep}{Chemistry}\n\n[Reads]\n{read1}\n{read2}\n\n'
-            '[Settings]\nReverseComplement{sep}{ReverseComplement}\n\n'
-            '[Data]\n{data}')
+            '[Settings]\nReverseComplement{sep}{ReverseComplement}\n%s'
+            '\n[Data]\n{data}' % (
```
wasade: Mixing dict formatting with string formatting is a bit confusing to look at, would it be possible to do this such that the optional string (based off of `is_amplicon`) is part of the formatting dict used when filling the template?

qiyunzhu: Good point! Let me edit.

qiyunzhu:
Hi @wasade @AmandaBirmingham The Travic CI build for my PR failed because of an Internet connection issue. I think by restarting it things will be resolved. But I don't know how to restart Travis CI building without modifying the commit. Can either of you help? Thanks!

wasade:
It doesn't seem like i have permissions to restart the build or its due to being on mobile

antgonza:
Already did ...

wasade:
Thanks!

qiyunzhu:
@antgonza Thank you!

Looks like something else went wrong, which isn't related to the code itself. Could you please take a look at the Travis message? Thank you!

antgonza:
No idea what's wrong, just restarted to be sure.

qiyunzhu:
@Thanks! HTTP error again. Perhaps it is just not a good day to run Travis CI...

antgonza:
Can be one of those days ...

antgonza:
It seems like the tested passed but this is still red due to @wasade's requested changes ...

labman/db/process.py
```diff
@@ -2778,7 +2778,7 @@ def _format_sample_sheet(self, data, sep=','):
         sample_sheet : str
             the sample sheet string
         """
-
+        is_amplicon = self.assay == 'Amplicon'
```
AmandaBirmingham: With this addition, there would be THREE places in process.py where we were explicitly checking whether the value of self.assay is equal to the string 'Amplicon' (generate_sample_sheet, generate_prep_information, and here).  I think we're just asking for bugs in which, in the future, we change one or two of them but forget to update them all.  I can't take building on this bad practice and want us to stop it here :)

The most minimal way I can think of to address this would be to place the string 'Amplicon' in a constant somewhere, and then have the comparisons (here and elsewhere) made against that constant.  I suppose we'd probably want to do the same for the string 'Metagenomics' that is (for now) the only other acceptable value of the assay attribute.   (The assay attribute is specifically associated with a SequencingProcess, so perhaps the constants could be associated with that.)   This isn't perfect as the assay values come out of the db, so someone could, theoretically, change them there so they mismatched with the code, but at least it minimizes the change points we'd have to touch in such a case.

qiyunzhu:
Hi @AmandaBirmingham I talked to Greg in person before I coded this PR. He told me that the i5 index columns are NOT needed. Are you sure we should add them back in?

AmandaBirmingham:
@qiyunzhu : Yes.  Per the comment above, we are trying to set up tests against real data to ensure we can re-generate real, known-good outputs, and those columns are in the known-good outputs.  (At least we know that, while they may not be needed, they can't be hurting anything :)

labman/db/process.py
qiyunzhu: Thanks for your thoughtful comments! I recall that, once I tried to modify the properties of a `SequencingProcess` object in `test_process.py`, and was blocked. So I guess this `assay` attribute is also safe from user modification. FYI.

AmandaBirmingham: It is good to know the users can't muck with that value!  I hadn't even thought about that nasty possibility; I'm worried about us *coders* deciding to change that string at some point and not remembering to search for all the instances :)

qiyunzhu:
@AmandaBirmingham Got it! Columns added.

AmandaBirmingham:
Ok, it does not look like the final change request is going to be addressed.  Therefore I am going to approve and merge this PR and then make that change myself.

qiyunzhu:
@AmandaBirmingham Thank you!

Issue #301, opened 2018-08-01
Update generation of Preparation Sheets contain correct columns/values


AmandaBirmingham:
Raised in amplicon walkthrough 07/30/2018
Note: below, "(empty)" means this column is intentionally left empty. "(empty)" is NOT part of the column name.

sample_name
BARCODE
PRIMER
Primer_Plate
Well_ID
Plating
ExtractionKit_lot
Extraction_robot
TM1000_8_tool
Primer_date
MasterMix_lot
Water_Lot
Processing_robot
TM300_8_tool
TM50_8_tool
Sample_Plate
Project_name
Orig_name
Well_description
EXPERIMENTAL_DESIGN_DESCRIPTION
LIBRARY_CONSTRUCTION_PROTOCOL
LINKER
PLATFORM
RUN_CENTER
RUN_DATE (empty)
RUN_PREFIX (empty)
pcr_primers
sequencing_meth
target_gene
target_subfragment
center_name
center_project_name (empty)
INSTRUMENT_MODEL (empty)
runid (empty)

AmandaBirmingham:
Hi @qiyunzhu : This issue will require modifying the `generate_prep_information` method in labman/db/process.py , specifically the part for Amplicon prep info generation.  You can see an example of what the current output looks like in the `TARGET_EXAMPLE` constant in labman/db/tests/test_process.py .  However, the wet lab indicates they are expecting it to have a different set of column names/order, etc than it currently creates.  I will email you the correct example directly.

qiyunzhu:
Hi @AmandaBirmingham I explored the header names and still wasn't sure how to start with. First, I wonder if there is a mapping between old and new names? I guess `MasterMix_lot` is identical to `master_mix`, but I don't know if `kingfisher_robot` and `Processing_robot` are the same thing. Second, the new list of header names is longer and new columns need to retrieved by modifying the SQL command. It will be great to know where they are stored in the database. Some seem to be hardcoded I guess (e.g., `target_gene` == `16S rRNA`). My limited understanding of SQL language and labman database structure may make this task bit challenging. So I probably will need lots of help from you...

qiyunzhu:
This is the best I can guess. Many columns are missing from the original report.

new | old | empty | uncertain
-- | -- | -- | --
sample_name | sample_name |   |  
BARCODE | primer_set_composition |   | ?
PRIMER |   |   |  
Primer_Plate |   |   |  
Well_ID | well |   |  
Plating |   |   |  
[...]

PR #319, opened 2018-08-14
SQL script changes/additions to populate production database

### LINKED ISSUE #262: Populate real labman db with needed real equipment/etc records

### LINKED ISSUE #288: Add 18S and ITS primer plates to Prepare Amplicon Libraries
in each plate's "Primer plate" section
Raised in amplicon walkthrough 07/30/2018

AmandaBirmingham:
Fixes #262 : Populate real labman db with needed real equipment/etc records by adding the new equipment types identified by the wet lab (but NOT new equipment, given decision that users should make use of the GUI for that), adding two real users (one wet lab, one developer) with admin access (who are responsible for adding other users through GUI), modifying description of plate types to not conflict with the terminology the wet lab uses.

Issue #262 also bleeds over into (at least) two other issues, so this PR fixes #288 Add 18S and ITS primer plates to Prepare Amplicon Libraries (which is a subset of 262 that was raised by a wet lab user) by adding the 18S and ITS primer set plate maps.  It also extends and fills in the descriptive info about the amplicon primer sets (e.g. linker primer sequence), which lays some necessary groundwork for #301 Update generation of Preparation Sheets contain correct columns/values because some of the info needed to generate the Prep Info files previously didn't actually exist anywhere in the db.


AmandaBirmingham:
@antgonza Ugh, you're very right.  I've regenerated the labman.html and labman.dbs files to reflect the table structure changes to marker_gene_primer_set, and updated the PR.

labman/db/tests/test_equipment.py
```diff
@@ -16,26 +17,26 @@
...
+               {'equipment_id': 19, 'external_id': 'IGM-HiSeq4000'},
+               {'equipment_id': 8, 'external_id': 'JER-E'},
                {'equipment_id': 11, 'external_id': 'KF1'},
                {'equipment_id': 12, 'external_id': 'KF2'},
                {'equipment_id': 13, 'external_id': 'KF3'},
                {'equipment_id': 14, 'external_id': 'KF4'},
-               {'equipment_id': 15, 'external_id': '108379Z'},
-               {'equipment_id': 16, 'external_id': '109375A'},
-               {'equipment_id': 17, 'external_id': '311411B'}]
+               {'equipment_id': 18, 'external_id': 'KL-MiSeq'},
```
wasade: Open question: do we record the chemistry version used (implicitly or explicitly)? If not, should we?

scripts/generate_18s_its_primer_sql
```diff
@@ -0,0 +1,1644 @@
...
+AATCTTGCGCCG	GTGTGCCAGCMGCCGCGGTAA	8	H4
+GGAAATCCCATC	GTGTGCCAGCMGCCGCGGTAA	8	H5
+GACCGTCAATAC	GTGTGCCAGCMGCCGCGGTAA	8	H6
+TTGGAACGGCTT	GTGTGCCAGCMGCCGCGGTAA	8	H7
+TCCTAGGTCCGA	GTGTGCCAGCMGCCGCGGTAA	8	H8
+TCCTCACTATCA	GTGTGCCAGCMGCCGCGGTAA	8	H9
+GCCTGCAGTACT	GTGTGCCAGCMGCCGCGGTAA	8	H10
+GCCCAAGTTCAC	GTGTGCCAGCMGCCGCGGTAA	8	H11
+ATAAAGAGGAGG	GTGTGCCAGCMGCCGCGGTAA	8	H12
+￿￿￿￿￿￿￿￿￿￿￿￿￿￿￿"""
```
wasade: Should these characters be here?

scripts/generate_18s_its_primer_sql
```diff
@@ -0,0 +1,1644 @@
...
+CCGCACTCAAGT	CGGCTGCGTTCTTCATCGATGC	8	H3
+TGTGGAAACTCC	CGGCTGCGTTCTTCATCGATGC	8	H4
+TTAGGCAGGTTC	CGGCTGCGTTCTTCATCGATGC	8	H5
+TAAGACTACTGG	CGGCTGCGTTCTTCATCGATGC	8	H6
+CGCGAAGTTTCA	CGGCTGCGTTCTTCATCGATGC	8	H7
+CGATACACTGCC	CGGCTGCGTTCTTCATCGATGC	8	H8
+TTGAAATCCCGG	CGGCTGCGTTCTTCATCGATGC	8	H9
+GTTAGGGAGCGA	CGGCTGCGTTCTTCATCGATGC	8	H10
+TTACTGTGGCCG	CGGCTGCGTTCTTCATCGATGC	8	H11
+ATATAAGGCCCA	CGGCTGCGTTCTTCATCGATGC	8	H12￿￿￿￿￿
```
wasade: Same re the extraneous characters

scripts/generate_18s_its_primer_sql
```diff
@@ -0,0 +1,1644 @@
...
+def generate_insert_values(last_plate_id, last_container_id):
+    set_definitions_by_name = {"EMP 18S": emp_18s_tsvs,
+                               "EMP ITS": emp_its_tsvs}
+    # from records created by db_patch_manual.sql
+    primer_set_ids_by_name = {"EMP 18S": 3, "EMP ITS": 4}
+
+    num_plates_in_set = 8
+    num_rows_on_plate = 8
+    num_cols_on_plate = 12
+    num_wells_on_plate = num_ro
```
wasade: I don't understand the hardcoded values here, would these ever cause a real problem in practice? It seems like this is a use once and burn it script, right?

wasade: Should any sanity checking be performed on the tsv data, such as uniqueness of barcodes, plate + well position, that all positions are represented, and that the barcodes decode as golay? I don't know if any of those tests would be extreme (e.g., we have a very high degree of confidence the inputs are perfect)

labman/db/tests/test_equipment.py
AmandaBirmingham: @wasade Can you educate me on what the "chemistry version" is?  I am not aware of anything called that being tracked in the db but can't be sure without understanding the concept :)

scripts/generate_18s_its_primer_sql
AmandaBirmingham: Hmm, I don't think they'd cause a problem but I'll refactored the script so they are default arguments to the function rather than hard-coded constants inside the function.

This IS basicaly a "use once and burn it" script, but I checked it into the repo (rather than just trashing it) because I figured that if we ever came up with ANOTHER primer set (and therefore needed to input it into the db) it would be easier to modify this script than to rewrite the same functionality again--which I think we would definitely end up doing, because writing the SQL to fill out all these tables by hand for multiple plates would be *at least* as error-prone as scripting it out.

AmandaBirmingham: For myself, I see testing the tsv inputs as outside the scope of this issue; I feel like it is reasonable to expect that the lab provide the correct barcode information.  The text in the inputs is copied directly from their provided xlsx files with zero modification by me (except to remove those weird trailing characters you pointed out above :) .

labman/db/tests/test_equipment.py
wasade: The chemistry on the instruments changes over time and is at least a possible source of technical bias. But Im not sure if that is being tracked. It sounds like it is not. Tagging people on my phone does not seem to be working at the moment, would it be possible to ask Austin for an opinion?

PR #321, opened 2018-08-15
Add handling for CRLF line endings to Pico Green quantification file handling

### LINKED ISSUE #337: Add handling for CRLF line endings to Pico Green quantification file handling
Argh, apparently some of the plate-readers produce files that end with CRLF (rather than just LF or just CR, as we saw in previously provided examples).

AmandaBirmingham:
Argh, apparently *some* of the plate-readers produce files that end with CRLF (rather than just LF or just CR, as we saw in previously provided examples).  The tests are longer than the actual fix :-/

EDIT: logged a bug for this.  fixes #337

### DIFF
--- a/labman/db/process.py
+++ b/labman/db/process.py
@@ -1757,7 +1757,11 @@ class QuantificationProcess(Process):
 
     @staticmethod
     def _rationalize_pico_csv_string(contents):
-        # Plate reader files end with CR; convert to LF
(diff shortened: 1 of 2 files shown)


tanaes:
in the past, these have shown up when the files get emailed because the
UCSD email system silently and automatically adds LFs onto the end of files
with a .csv extension. Fun!

PR #328, opened 2018-08-17
fix #301

### LINKED ISSUE #301: Update generation of Preparation Sheets contain correct columns/values

antgonza:


AmandaBirmingham:
Hi @antgonza : the timestamp change has just been merged into master :) so could you please integrate that into this PR and address any check failures that remain afterwards?  Would love to review and merge this by the end of the week!

antgonza:
Wait on review, need to do another pass ...

antgonza:
Just restarted the build cause locally the error is random, not sure if related to other issues but it seems like the plate number changes from 2-4 in some occasions, @AmandaBirmingham any ideas? 

```
AssertionError: 'samp[661 chars]late 2\tCannabis Soils\t1.SKB1.640202\tTest pl[40336 chars]\t\n' != 'samp[661 chars]late 4\tCannabis Soils\t1.SKB1.640202\tTest pl[40336 chars]\t\n'
```

AmandaBirmingham:
Yep, I strongly suspect I know the answer to this one.  Having looked through the code, I believe that the `ORDER BY study_id, sample_id, row_num, col_num` clause in the big sql statement in `generate_sample_sheet` is not adequately constrained to ensure a deterministic output.  Within the same study, the *exact same* sample_id (with the exact same row_num and col_num) can appear on *multiple different plates*, and without more constraint, which one is returned first will be random.  There are apparently a lot of situations like this in the codebase, which were masked previous to  the merge of PR 331 (Corrections of and expansions to unit test database) because all the tests were written using only a single test plate :(

**However**, that is not the only issue with this sql statement :), because I see that the known-good prep file does not order samples by sample id at all but rather by plate and then by position (first by column, then by row--the opposite of what the above sql does):

screen shot 2018-08-24 at 9 53 25 am

Since the goal is to reproduce the known-good file, `generate_sample_sheet` should follow this order.

antgonza:
Thank and sounds good, any suggestions on how to adequately constrain the SQL?

AmandaBirmingham:
That said, if you decide you want to open a new issue for the plate/column/row sorting instead of handling it in this PR, I'm ok with that.  In that case, it would still be necessary to add an additional constraint to the sql ORDER BY statement, or else to explicitly sort the results in python afterwards, so that the output order becomes deterministic and is thus stable every time we run the tests :)

AmandaBirmingham:
Ooops, sorry, our messages crossed ...

AmandaBirmingham:
My initial guess would be that ordering by plate, then col_num, then row_num would do it.  I don't know if there is actually supposed to be a study ordering ... the known-good file I have all comes from a single study, so I'd have to check that with the wet lab.

wasade:
Adding in the plate to ordering should do it, assuming the plate number is globally unique (to labman). 

Organizing the output by study _could_ be nice for subsequent debugging and human eyes. But, I can also see the value in having it ordered by physical location too.

AmandaBirmingham:
I think we will want to use the plate external_id (the user-assigned plate id), which I think is what @antgonza already has in `plate` and  *is* constrained to be unique within labman during the table set-up in db_patch.sql: 

`CONSTRAINT idx_plate UNIQUE ( external_id )`

I've reached out to Gail and the wet lab about whether they want additional ordering by study, but until/unless they say they do, I say we stick with straight-up positional ordering for now.

antgonza:
For consistency, should we do sample_plate, primer_plate, col_num, row_num or do you have any other suggestions?

antgonza:
Note that the sorting in SQL is ignored once everything moves to a dataframe, so decided to remove that sorting and use the pandas sorting ...

AmandaBirmingham:
Ok, so the latest test failure says:

screen shot 2018-08-24 at 2 06 25 pm

I began debugging by simply pulling your test's `TARGET_EXAMPLE` "known-good" string into excel so I could read it more easily, and it seems like there are some quite weird things going on in it:

screen shot 2018-08-24 at 2 11 51 pm

For example, it looks like:
1) It is not sorted by column and then by row within plate, but rather by well id--which leads to G10 coming after G1, etc.
2) There are only 24 rows for the first plate, which should have 95 (not 96 because H12 of each 96-well plate in the test db is empty).

Given this, I think the best place to start fixing this test failure is probably returning to the `TARGET_EXAMPLE` string and ensuring that it matches what you really think the code should be producing given the specs and discussion so far; for the info in `TARGET_EXAMPLE` currently, I'd be worried if this test WAS passing :)

labman/db/process.py
```diff
@@ -3090,23 +3090,66 @@ def generate_prep_information(self):
...
+                    -- TM50_8_tool
+                    epmotion_tm50_8_tool_id,
+                    -- sample_plate
+                    p1.external_id AS sample_plate,
+                    -- project_name
+                    study_alias AS project_name,
+                    -- orig_name
+                    sample_id AS orig_name,
+                    -- experiment_design_description
+                    '' AS experiment_design_description,
```
wasade: The design description is always empty?

labman/db/process.py
```diff
@@ -3090,23 +3090,66 @@ def generate_prep_information(self):
...
+                    -- extractionkit_lot
+                    extraction_kit_id,
+                    -- extraction_robot, both the epmotion robot and the
+                    -- kingfisher robot
+                    lpp.epmotion_robot_id AS lepmotion_robot_id,
+                    kingfisher_robot_id,
+                    -- TM1000_8_tool
+                    epmotion_tool_id,
+                    -- mAStermix_lot
+                    mASter_mix_id,
```
wasade: should the AS be capitalized?

labman/db/process.py
wasade: ...same question for some of those below

labman/db/process.py
```diff
@@ -3090,23 +3090,66 @@ def generate_prep_information(self):
...
+                    -- sample_plate
+                    p1.external_id AS sample_plate,
+                    -- project_name
+                    study_alias AS project_name,
+                    -- orig_name
+                    sample_id AS orig_name,
+                    -- experiment_design_description
+                    '' AS experiment_design_description,
+                    -- run_center
+                    'UCSDMI' AS run_center,
```
wasade: Is the center the same if we run on the miseq that sits in our lab space vs. the instruments that are part of the sequencing core? Or those at rady?

labman/db/process.py
```diff
@@ -3090,23 +3090,66 @@ def generate_prep_information(self):
...
+                    -- primer_date
+                    run_date AS primer_date,
+                    -- run_date
+                    '' AS run_date,
+                    -- RUN_PREFIX
+                    '' AS run_prefix,
+                    -- sequencing_meth
+                    'Sequencing by synthesis' AS sequencing_meth,
```
wasade: Would this be accurate if we were sequencing on Nanopore?

labman/db/process.py
```diff
@@ -3237,32 +3280,32 @@ def generate_prep_information(self):
...
                        USING (equipment_type_id)""")
-            equipment = {}
-            for row in TRN.execute_fetchindex():
-                row = dict(row)
-                eid = row.pop('equipment_id')
-                equipment[eid] = row
-
-            # and the reagents
+            equipment = {dict(row)['equipment_id']: dict(row)
+                         for row in TRN.execute_fetchindex()}
```
wasade: is it assured that `equipment_id` will be unique from this query?

labman/db/process.py
```diff
@@ -3237,32 +3280,32 @@ def generate_prep_information(self):
...
                        FROM labman.reagent_composition
                        LEFT JOIN labman.reagent_composition_type
                        USING (reagent_composition_type_id)""")
-            reagent = {}
-            for row in TRN.execute_fetchindex():
-                row = dict(row)
-                rid = row.pop('reagent_composition_id')
-                reagent[rid] = row
+            reagent = {dict(row)['reagent_composition_id']: dict(row)
+                       for row in TRN.execute_fetchindex()}
```
wasade: same q

labman/db/process.py
```diff
@@ -3237,32 +3280,32 @@ def generate_prep_information(self):
...
+            # 3/3. marker gene primer sets
+            TRN.add("""SELECT marker_gene_primer_set_id, primer_set_id,
+                           target_gene, target_subfragment, linker_sequence,
+                           fwd_primer_sequence, rev_primer_sequence, region
+                       FROM labman.marker_gene_primer_set""")
+            marker_gene_primer_set = {dict(row)['primer_set_id']: dict(row)
```
wasade: same q

wasade: ...to rephrase, I'm concerned about the following type of scenario:

```python
>>> {k: i for i, k in enumerate([1,1,1,2,2,2,3,3,3,4,4,4])}
{1: 2, 2: 5, 3: 8, 4: 11}
```

I don't have the schema well in mind, so it's not clear from just looking at these queries whether the above situation is or isn't possible

labman/db/process.py
```diff
@@ -3293,14 +3336,37 @@ def generate_prep_information(self):
                     result[nk] = val
 
                 # format some final fields
-                result['platform'] = equipment[
-                    result.pop('platform_id')]['description']
+                result['platform'] = 'Illumina'
```
wasade: This also seems surprising since we do have Nanopores in the lab, and it's of course possible other sequencing platforms will be used in the future

labman/db/process.py
```diff
@@ -3293,14 +3336,37 @@ def generate_prep_information(self):
...
+                        mgps['linker_sequence'], mgps['fwd_primer_sequence'])
+                    result['pcr_primers'] = 'FWD:%s; REV:%s' % (
+                        mgps['fwd_primer_sequence'],
+                        mgps['rev_primer_sequence'])
+                    result['linker'] = mgps['linker_sequence']
+                    result['target_gene'] = mgps['target_gene']
+                    result['target_subfragment'] = mgps['target_subfragment']
+                    result['library_construction_protocol'] = (
+                        'Illumina EMP protocol {0} amplification of {1}'
+                        ' {2}'.format(mgps['region'], mgps['target_gene'],
```
wasade: What is the region here? target_gene and target_subfragment make sense, however I don't believe the EMP protocol allows for variation in the subfragment of a given gene. 

Is it possible that the fwd/rev primers set on like 3352 will contradict the library construction protocol?

labman/db/process.py
AmandaBirmingham: Actually, this one is a mistake on my part--thanks for noticing it!  I classed this in the "leave empty" category (see below) because it is empty in the known-good example I have, but going back to my notes, I see that Gail suggested that we should pull it from the Qiita study description. 

Gail instructed us to always leave the following fields empty:
RUN_DATE
RUN_PREFIX
center_project_name
INSTRUMENT_MODEL
RUNID

I assume that someday in the future, labman or qiita or something will be expanded to close the loop by filling in these values after the sequencing is done ...

labman/db/process.py
AmandaBirmingham: When I asked "Is this always UCSDMI ?", the answer I got from Gail was "For our lab YES".  I interpreted that to mean that it was true for all prep info sheets created in the lab; should I be asking her a more nuanced question here, and if so, what?

labman/db/process.py
AmandaBirmingham: No, if it were another sequencing platform it would be different.  However, the reference sheet I have from the lab/Gail doesn't include a description of what it should be for Nanopore.  I can ask about that.

labman/db/process.py
wasade: Okay, thanks! Some of this is out-of-scope for labman too (e.g., the actual sequencing date)

labman/db/process.py
wasade: Okay. I think it would be useful to ask Gail if UCSDMI encompasses any sequencing instrument associated with the UCSD campus. If yes, than we're good. If no, than more granularity may be warranted

labman/db/process.py
wasade: Okay, thanks. I don't think questions on the platform should block soft launch, but we do have the capability to run on other platforms. I'm not sure if labman is supposed to be illumina specific.

labman/db/process.py
AmandaBirmingham: I'm surprised in the other direction, since here's the exchange I had when gathering specs for this field:
"Is this always Illumina?"
"Yes we don't have another platform we process although there are others in Qiita"

If we DO need to make some decision by sequencing technology, we will need to store that info in the labman database, and it is not there right now--for example, the "description" field that was previously referenced holds the human-readable NAME of the equipment type being used ("MiSeq", "HiSeq3000", etc) not the sequencing technology.

labman/db/process.py
AmandaBirmingham: The field called "region" in the database holds the region of the gene that is amplified by the actual, particular primer pair in use; its values were provided by Gail and are:
16S V4: "515fbc, 806r"
18S V9: "1391f/EukBr"
ITS 1/2: "ITS1fbc, ITS2r"

The only way I can think of that the primer sets could contradict the library construction protocol was if we linked the primer records in the db to the wrong marker gene primer set; this is the one-time set-up done in db_patch_manual.sql and populate_prod_db.sql, so hopefully we do it right once and we're golden :)

wasade: Ah okay, so the wording is "Illumina EMP protocol 515fbc, 806r amplication of 16S V4" or to that effect. 

And thanks for the clarification on the lack of a possible contradiction :)

AmandaBirmingham:
BTW, any @wasade comments I didn't reply to I'm leaving to @antgonza because I don't have their answers :)

labman/db/process.py
AmandaBirmingham: Ok, I've added #342 Support generating prep info files for non-Illumina sequencing runs

labman/db/process.py
AmandaBirmingham: Now incorporated into #342 Support generating prep info files for non-Illumina sequencing runs

labman/db/process.py
AmandaBirmingham: Ok, I've put a question out to Gail.  If I learn that we need to make the behavior for this field more nuanced, I'll share that here.

labman/db/process.py
```diff
@@ -3293,14 +3336,37 @@ def generate_prep_information(self):
...
+                                      mgps['target_subfragment']))
 
                 if sid is not None and study_id is not None:
                     study = Study(study_id)
                     if study not in data:
                         data[study] = {}
-                    data[study][content] = result
+                    # if we want the sample_name.well_id, just replace sid
+                    # for content
+                    data[study][sid] = result
```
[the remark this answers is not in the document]
AmandaBirmingham: I also checked this one with Gail.  She confirms that:
1) The sample_name column must exactly match what is in Qiita (so we can't use sample content there, even if there are duplicates) and
2) Every sample that is in the run must be represented in the prep info file (so, even if there are duplicates, we can't leave those out of the file)
        * Note that while duplicate samples on the same plate or even in the same run is a very undesirable case in the real world, *it is the norm in the test data*.  If you want to change the test db population and attendant unit tests so that they represent a more realistic case (no duplicate samples), that would be great!

In the case of duplicate samples in the same run, Gail manually goes into Qiita and adds a new sample (to represent the duplicate) and then manually modifies the name of the duplicate in the prep file to match the name she created for the new sample in Qiita.  She has asked for Labman to show a warning when users download a prep info file that has duplicate samples in it, telling them that they will need to go through this process.  I have added this as #343 Warn user if prep info file contains duplicate samples.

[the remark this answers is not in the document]
antgonza: Kewl, just note that in the current state we will have a difference in the blanks and spikes in the prep and the sample information file, which we will need to add to the sample info in Qiita somehow.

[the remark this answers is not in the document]
AmandaBirmingham: @antgonza You are absolutely right that the foregoing applies only to experimental samples--thanks for pointing that out!  The known-good example we have for the prep info file shows that blanks are qualified by their well id while samples are not:

screen shot 2018-08-28 at 10 49 49 am

Furthermore, the `labman.sample_composition` table doesn't even store a `sample_id` for blanks and other controls:

screen shot 2018-08-28 at 10 52 03 am

Therefore, it looks like the `sample_name` column in the prep info file will need to contain the `sample_id` for experimental samples (i.e., those with `sample_composition_type_id`=1) and the `content` for everything else.  Good catch!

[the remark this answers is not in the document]
AmandaBirmingham: Related to @antgonza 's point above about blanks, etc: it looks like handling them requires a modification to our way of generating the Well_Description, too.  The known-good file we got from the lab shows well descriptions like this:

screen shot 2018-08-28 at 11 04 35 am

However, since as noted above the `sample_composition` table doesn't even store a `sample_id` for blanks and controls, following our approach of building the well description from the `sample_id` in these cases gives well descriptions with `None` in them, like this:

screen shot 2018-08-28 at 11 05 05 am

I'm coming to the conclusion that maybe *everywhere* in the prep info file generation that we incorporate `sample_id` for experimental samples (e.g., in sample name, well description, original name, more?), we need to instead incorporate `content` for the non-experimental samples to match the known-good.   @antgonza , what do you think?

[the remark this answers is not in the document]
antgonza: I think this should be a case by case scenario:
- orig_name = name from qiita or empty for blanks/spikes
- sample_name = original name for unique samples, if samples are duplicated, use content; for blanks/spikes use content
- rest, like well description: use the previous sample name

This is an easy change in the current code; however, not easy to test as we will need to create a test with only unique sample ids and I have _no_ idea of how to do that ...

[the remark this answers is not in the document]
AmandaBirmingham: From the known-good we got from the lab, it looks like they currently expect orig_name to match sample_name even for blanks, etc (see below) so let's follow their convention on that.  I agree with the suggestions for all the rest!

screen shot 2018-08-28 at 11 28 29 am

Re testing: Yeah, testing is going to be sticky :(  The labman test db only knows about ONE Qiita study ("Identification of the Microbiomes for Cannabis Soils") and says that study only has 27 samples (not enough to make a full plate without either repetition or lots of blanks).  It looks like this info comes from qiita/qiita_db/support_files/populate_test_db.sql , rather than from anything within labman, so I'm hesitant to try to modify its content in the labman db population scripts.   I don't want to change the existing test plates in the db to not have any duplicate samples because we need to be able to test the behavior of full plates.  Alternatively, adding new test plates that don't have duplicate samples on them would be a big effort in expanding the db population script; that effort should really be put in at some point, but it is going to be enough of a pain that I don't want to do it until such time as we have access to a larger test study in the qiita tables to "plate".

Given all this, I think the best of a host of bad options may be to change the db population script so that there are a *small number* of unique samples on the existing plates.  This will be a pain and kind of hacky (and will require some changes to the existing unit tests--ugh!) but will mean that we will be able to test all the use cases (unique experimental samples, duplicated experimental samples, and blanks/etc) at least once.  I can probably get a PR ready for that today.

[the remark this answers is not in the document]
antgonza: Note that it's pretty easy to create new studies in qiita via the qiita-api or even adding new samples to an existing study without having to modify qiita itself but this will need to be done via the qiita_client (which is part of of the [db reset for tests](https://github.com/jdereus/labman/blob/master/labman/db/testing.py#L36-L39)).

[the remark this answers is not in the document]
AmandaBirmingham: A PR that puts one unique sample on each of the sample plates in the test db (replacing one of the samples that was previously a repeat) is now available: #346 Test database extension: some unique samples

labman/db/process.py
wasade: Thanks!

labman/db/process.py
wasade: Thanks

labman/db/process.py
wasade: Thanks

labman/db/process.py
wasade: Thanks

labman/db/process.py
wasade: Thanks

antgonza:
@AmandaBirmingham, just wondering if there is anything missing from this PR? Thanks!

AmandaBirmingham:
@antgonza It looks to me like we are still getting Nones incorporated in the Well_description for blanks (see example in https://github.com/jdereus/labman/pull/328#discussion_r213419324 ); as far as I know, that's the last remaining item!

antgonza:
Good catch, fixing.

Issue #329, opened 2018-08-17
Change dates from time-zone-aware timestamp to naive timestamp?


AmandaBirmingham:
@wasade @tanaes @ElDeveloper please weigh in:

Currently the labman database stores time-zone aware dates, in order to be technically able to differentiate potentially different times (e.g., 2017-10-25 19:10:25 in New York is not the same time as 2017-10-25 19:10:25 in Beijing).  However, this incurs additional steps when unit-testing string outputs containing dates and times, as the strings differ based on the time zone in which the test is run and so have to be generated within the test rather than hard-coded (see an example below.)

@antgonza suggests changing the date fields in the database from timezone-aware to timezone-naive (timestamptz to timestamp), since labman is not anticipated to be run in multiple locations.  Shall we?

https://github.com/jdereus/labman/blob/ba272c7ab0e197c094e24a9630c64dc6f72da9ea/labman/db/tests/test_process.py#L1496-L1510

antgonza:
As you guys (@wasade @tanaes @ElDeveloper) can imagine this is blocking finishing the solution for #301 so it will be great to hear from you. 

Now, I was working on the changes to make this happen - so we could have a better discussion point - and realized a new thing: [get_date_format](https://github.com/jdereus/labman/blob/master/labman/db/process.py#L88) is format as: `'%Y-%m-%d %H:%M'` but all the tests are like `'%Y-%m-%d %H:%M:%S'`, which has this not fun implication:
```
>>> from datetime import datetime
>>> datetime.strptime('2018-01-01 00:00:00', '%Y-%m-%d %H:%M:%S')
datetime.datetime(2018, 1, 1, 0, 0)
>>> datetime.strptime('2018-01-01 00:00:00', '%Y-%m-%d %H:%M')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/antoniog/miniconda3/envs/labman/lib/python3.5/_strptime.py", line 510, in _strptime_datetime
    tt, fraction = _strptime(data_string, format)
  File "/Users/antoniog/miniconda3/envs/labman/lib/python3.5/_strptime.py", line 346, in _strptime
    data_string[found.end():])
ValueError: unconverted data remains: :00
>>> datetime.strptime('2018-01-01 00:00', '%Y-%m-%d %H:%M:%S')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
[...]

PR #330, opened 2018-08-18
rm timezone timestamps

antgonza:
To complement discussion on: https://github.com/jdereus/labman/issues/329


labman/db/tests/test_process.py
```diff
@@ -40,21 +40,11 @@ def _help_compare_timestamps(input_datetime):
...
     is_close = time_diff.total_seconds() < 60
     return is_close
 
 
-def _help_make_datetime(input_datetime_str):
-    # input_datetime_str should be in format '2017-10-25 19:10:25-0700'
-    return datetime.strptime(input_datetime_str, '%Y-%m-%d %H:%M:%S%z')
-
-
-def _help_format_datetime(input_datetime):
```
AmandaBirmingham: I think we may want to keep this: it is much easier than writing out `datetime.strftime(input_datetime, Process.get_date_format())` each time we need to test a date, and using `str(input_datetime)` (while shorter) undesirably includes the seconds.

labman/db/tests/test_process.py
```diff
@@ -187,8 +176,7 @@ def test_notes(self):
 class TestReagentCreationProcess(LabmanTestCase):
     def test_attributes(self):
         tester = ReagentCreationProcess(5)
-        self.assertEqual(tester.date,
-                         _help_make_datetime('2017-10-23 09:10:25-0700'))
+        self.assertEqual(str(tester.date), '2017-10-23 09:10:25')
```
AmandaBirmingham: If we want to move to testing the dates as *strings*, then I think all the changed code instances like this one, here and below, are impacted by above comments re removing seconds from expected string values of the dates and re possibly keeping `_help_format_datetime`. 

Alternatively, we can keep testing them as date objects, and just change `_help_make_datetime` to strip and/or ignore the timezone information when doing the conversion to an object.

labman/db/process.py
```diff
@@ -86,12 +86,12 @@ def factory(process_id):
 
     @staticmethod
     def get_date_format():
-        return '%Y-%m-%d %H:%M'
+        return '%Y-%m-%d %H:%M:%S'
```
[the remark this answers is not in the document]
antgonza: K, not sure on the implication on tests and duplicated plate names ... but we'll figure it out.

labman/db/support_files/db_patch.sql
```diff
@@ -127,7 +127,7 @@ COMMENT ON COLUMN labman.plate.external_id IS 'Must be unique';
 CREATE TABLE labman.process (
 	process_id           bigserial  NOT NULL,
 	process_type_id      integer  NOT NULL,
-	run_date             timestamptz  NOT NULL,
+	run_date             timestamp  NOT NULL,
```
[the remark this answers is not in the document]
antgonza: k

labman/db/tests/test_plate.py
```diff
@@ -378,11 +378,11 @@ def test_properties(self):
         tester2 = Plate(26)
         self.assertEqual(len(tester2.quantification_processes), 2)
         self.assertEqual(tester2.quantification_processes[0].date,
-                         datetime.strptime("2017-10-25 19:10:25-0700",
-                                           '%Y-%m-%d %H:%M:%S%z'))
+                         datetime.strptime(
+                            "2017-10-25 19:10:25", Process.get_date_format()))
```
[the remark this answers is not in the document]
antgonza: yup 😭

labman/db/tests/test_process.py
antgonza: I actually like writing the full thing, cause is a  single line (AKA IMOO not function worthy) and we are talking on saving chars vs. lines, and can yield the confusion that we had between _help_format_datetime vs _help_make_datetime ... but fine ...

labman/db/tests/test_process.py
antgonza: ... k ...

labman/db/tests/test_process.py
AmandaBirmingham: Ok; I'm willing to cede this one as a personal preference rather than a necessity :D

labman/db/tests/tester.py
```diff
@@ -215,20 +215,24 @@ def integration_tests():
...
         raise ValueError(
             'Amplicon sample sheet does not match expected regex:\n%s' % obs)
 
-    obs = shotgun_seq_process.generate_sample_sheet()
-    res = re.match(EXP_SHOTGUN_SAMPLE_SHEET, obs)
-    if res is None:
-        raise ValueError(
-            'Shotgun sample sheet does not match expected regex:\n%s' % obs)
+    # ToDo: #327
+    # shotgun_seq_process = integration_tests_shotgun_workflow(user, samples)
```
wasade: should the commented code remain?

labman/gui/templates/sequence_run_list.html
```diff
@@ -15,11 +15,15 @@
...
           "<span class='glyphicon glyphicon-download'></span> " +
           "Download Sample Sheet</a>";
-        var preparationSheets = "<a href='/process/sequencing/" + row[5] +
-          "/preparation_sheets' class='btn btn-success'>" +
-          "<span class='glyphicon glyphicon-download'></span> " +
-          "Download Preparation Sheets</a>";
 
+        // ToDo: #327
+        var preparationSheets = '';
+        if (row[3] == 'Amplicon') {
```
wasade: would it make sense to assign variable names for `row[3]` and `row[5]` to avoid magic numbers?

PR #339, opened 2018-08-23
Closes #329 change from timestamptz to timestamp for process dates

### LINKED ISSUE #326: Disable Preparation Sheet download for shotgun sequencing runs
Currently Labman offers the user the ability to download a "Preparation Sheet" (aka prep info file) for shotgun experiments (see middle entry in list below):

screen shot 2018-08-16 at 3 55 21 pm

However, per Rodolfo, "We [the wet lab] don't generate a prep info file for metagenomics"; this is currently done by Gail, and the format/contents of the file that Labman currently creates does not match that of the file that Gail produces.  

Since the wet lab does not currently create prep info files for shotgun runs, creating them in Labman is not a requirement for soft launch; the ability to create these for shotgun runs should be removed from the interface until such time as the files are generated to meet Gail's specifications.

### LINKED ISSUE #329: Change dates from time-zone-aware timestamp to naive timestamp?

### LINKED ISSUE #336: Apply Process.get_date_format() to dates returned by GDNAExtractionProcessHandler.get and LibraryPrep16SProcessHandler.get
@antgonza noticed that there are still instances of dates that are being string-formatted with local format specifications rather than Process.get_date_format() (and are using a format specification *different* than that from get_date_format ):

https://github.com/jdereus/labman/blob/c2550ff94eda3116a631f548203c7d5ee24d3b2a/labman/gui/handlers/process_handlers/gdna_extraction_process.py#L50

https://github.com/jdereus/labman/blob/3255ee8dff553ee6642a50dbf6f960fd4802bb4d/labman/gui/handlers/process_handlers/library_prep_16s_process.py#L50

AmandaBirmingham:
Migrated from PR 330, work by @antgonza  :
closes #329 change from timestamptz to timestamp for process dates
fixes #336 ensure Process.get_date_format() is used for date display
fixes #326 disable preparation sheet download for shotgun runs

### DIFF
--- a/labman/db/tests/tester.py
+++ b/labman/db/tests/tester.py
@@ -223,6 +223,13 @@ def integration_tests():
         raise ValueError(
             'Amplicon sample sheet does not match expected regex:\n%s' % obs)
 
+    # shotgun_seq_process = integration_tests_shotgun_workflow(user, samples)
+    # obs = shotgun_seq_process.generate_sample_sheet()
+    # res = re.match(EXP_SHOTGUN_SAMPLE_SHEET, obs)
+    # if res is None:
+    #     raise ValueError(
+    #         'Shotgun sample sheet does not match expected regex:\n%s' % obs)
+
 
 def stress_tests_amplicon_workflow(user, samples, num_plates=1):
     print('Amplicon workflow', flush=True)
--- a/labman/db/tests/tester.py
+++ b/labman/db/tests/tester.py
@@ -216,19 +216,12 @@ def integration_tests():
     samples = get_samples()
     user = User('test@foo.bar')
     amplicon_seq_process = integration_tests_amplicon_workflow(user, samples)
-    shotgun_seq_process = integration_tests_shotgun_workflow(user, samples)
 
     obs = amplicon_seq_process.generate_sample_sheet()
     res = re.match(EXP_AMPLICON_SAMPLE_SHEET, obs)
     if res is None:
         raise ValueError(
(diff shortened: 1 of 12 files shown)


labman/db/tests/tester.py
```diff
@@ -216,20 +216,13 @@ def integration_tests():
     samples = get_samples()
```
antgonza: Just want to be sure that you want to delete vs. comment out these lines ... we are gonna need them back some time soon ...

AmandaBirmingham: Fair point!  I've put the lines to run the comparison back in but not the actual regex against which the comparison is done, as that will have to change pretty much from the ground up when we get around to bringing this feature back.

Issue #323, opened 2018-08-16
Fix "reformat" option in Normalize gDNA Plates


AmandaBirmingham:
According to Rodolfo, what this is supposed to do is: " eformat from fully interweaved to quadrants by columns (Plate_1 : cols 1-6, Plate_2 : cols 7-12, Plate_3 : cols 13-18, Plate_4 : cols 19-24. We rarely use this function and we might want to come up with another solution but it was written for when we wanted to run a subset of the plates. Let’s say that for some random reason we only wanted to run plates 1 & 4, if we left them fully interweaved, we would have a lot of empty wells in the plate that would result in wasted tips in the mosquito (the mosquito does transfers in full columns). However, if we reformat the plate to fully occupy columns, we would get rid of that tip waste."

However, it doesn't appear that this is what the code currently does; when tried on a full plate, it leave half a column blank while putting two samples per well into the other half:
<img width="399" alt="screen shot 2018-08-14 at 9 49 38 am" src="https://user-images.githubusercontent.com/10677935/44187128-40c40f80-a0d1-11e8-972c-a1fbe1f99970.png">

After this is addressed, the reformat option should be put back on the Normalize gDNA Plates page.

charles-cowart:
This should be resolved before full launch.

wasade:
It doesn't seem there are unit tests in place for the reformat. The offending code for the issue is I believe [here](https://github.com/jdereus/labman/blob/master/labman/db/process.py#L1166-L1174). Just noting to help reduce the search process when tackling this issue.

During investigation, a potential bug was observed surrounding the adding of plates on this page, now noted in #458.

charles-cowart:
Thanks! (Note link is now here: https://github.com/jdereus/labman/blob/master/labcontrol/db/process.py#L1166-L1174)

PR #340, opened 2018-08-25
Correct order of i5/i7 primer assignment and remove reformat option

### LINKED ISSUE #322: Remove "Reformat" option from Normalize gDNA Plates page
The reformatting functionality appears to be broken.  The wet lab (in the person of Rodolfo) says it isn't necessary for soft launch:
"I feel like we will increase LabMan’s features after it’s stable doing ordinary Kapa Hyper Plus runs. 

I don’t think reformatting compressed 384 well plates from interleaved to quadrants is necessary for soft launch. "

However, since it doesn't work, it has to be removed from the interface so no one uses it accidentally before it gets fixed.

### LINKED ISSUE #335: Correct order of i5/i7 primer assignment in LibraryPrepShotgunProcess.create
The LibraryPrepShotgunProcess.create method assigns primer combos by reading across each well of the plate, as shown below:
A1: combo 1
A2: combo 2
A3: combo 3
A4: combo 4
...
B1: combo 13
... etc

(Note that this is NOT the same as the order in which the primer combos are assigned to the one shotgun library prep plate in the test database, which is used as the basis of many tests; see issue #334 ).

According to known-good examples provided by the wet-lab, shotgun primer combos should be assigned in an interleaved fashion, as though reading through each of the original four plates that were compressed onto the library prep plate, column by column, starting with the top-left plate and then going to top-right, bottom-left and bottom-right:
A1: combo 1
C1: combo 2
E1: combo 3
....
A3: combo 9
C3: combo 10
...
A2: combo 97
C2: combo 98
...
...
B1: combo 183
B3: combo 184
... etc

AmandaBirmingham:
fixes #335 Correct order of i5/i7 primer assignment in LibraryPrepShotgunProcess.create
fixes #322 Remove "Reformat" option from Normalize gDNA Plates page


labman/db/process.py
```diff
@@ -640,23 +641,98 @@ class GDNAPlateCompressionProcess(Process):
...
+            Number of columns on large plate (e.g., 24 for a 384-well plate);
+            must be even.
+
+        Yields
+        -------
+        InterleavedPosition namedtuple
+
+        Raises
+        ------
+        ValueError
```
antgonza: My _guess_ is that there is some kind of relationship between of num_quarters and total_num_rows/cols. In other words, I don't think is possible to have num_quarters = 1 and total_num_rows = total_num_cols = 0; which I _think_ are valid values for these tests, right? Thus, I guess we need to add other kind of tests to validate that relationship.

labman/db/process.py
```diff
@@ -1379,19 +1473,60 @@ def create(cls, user, plate, plate_name, kappa_hyper_plus_kit, stub_lot,
...
+                    # in the i5 and i7 plate maps have separate database
+                    # entities from the positions of those primers on the
+                    # actual physical primer working plates being used, the
+                    # primers are at the same positions (of course!) in the
+                    # working plates made from a given primer set plate map
+                    # as they are in the plate map itself--duh :)
+                    # Thus, we use the position of the relevant i5 and i7
+                    # primers in the plate map (gotten above) to find the
+                    # real, physical compositions for the primers at those
+                    # positions on the real, physical primer working plates.
```
antgonza: Why the -1? Perhaps worth adding to the comments? Also, is this an expensive step? Just wondering if better to load all the i5-7 compositions before starting this and just use that info ...

labman/db/process.py
```diff
@@ -1379,19 +1473,60 @@ def create(cls, user, plate, plate_name, kappa_hyper_plus_kit, stub_lot,
...
+                    i5_comp = i5_layout[i5_well.row - 1][i5_well.column - 1]\
+                        .composition
+                    i7_comp = i7_layout[i7_well.row - 1][i7_well.column - 1]\
+                        .composition
+
+                    # note adding 1 to the row/col from p, as those are
```
antgonza: Sounds messy ... worth improving? Mainly wondering if we need to open an issue about it ...

labman/db/tests/test_process.py
```diff
@@ -988,12 +988,16 @@ def test_create(self):
 
         plates = obs.plates
         self.assertEqual(len(plates), 1)
-        layout = plates[0].layout
-        self.assertEqual(layout[0][0].composition.i5_composition,
-                         PrimerComposition(1523))
-        self.assertEqual(layout[0][0].composition.i7_composition,
-                         PrimerComposition(1524))
-        self.assertIsNone(layout[-1][-1])
+        obs_primer_layout = [[[well.well_id if well is not None else None,
```
antgonza: 😒 this tests is kind of odd, right? Basically here we are generating obs_primer_layout within the test vs. in a single function that generates this result ... this makes me wonder if we should be testing like this or some other way ...

labman/gui/handlers/process_handlers/test/test_gdna_compression_process.py
```diff
@@ -35,7 +35,7 @@ def test_get_gdna_plate_compression_process_handler(self):
         self.assertEqual(response.code, 404)
 
     def test_post_gdna_plate_compression_process_handler(self):
-        data = {'plates': json_encode(['22', '22']),
+        data = {'plates': json_encode(['22', '28']),
```
antgonza: Why the change from 22 to 28?

labman/gui/templates/normalization.html
```diff
@@ -276,7 +276,8 @@ <h3>Normalization</h3>
 </div>
 
 <!-- Reformat -->
-<div class="form-check" style='padding-bottom:10px'>
+<!-- display set to none (to hide from interface) until reformatting is fixed -->
```
antgonza: is there an issue for this? worth adding here?

labman/db/process.py
AmandaBirmingham: I'll add a comment on the -1, thanks!  My inclination is to hold off on performance optimizations at this stage unless we specifically notice they cause problems during test (e.g., getting plate.layout within the loop causes tests to fail with a time-out, so that was explicitly addressed) ... frankly, I suspect we'll get plenty of feedback from the wet-lab once this goes into test about which parts are slow, at which point we'll have to decide if we can address them piece-meal or need to do something more sweeping.

labman/db/process.py
AmandaBirmingham: Can you clarify more what part sounds messy?  I'm having trouble figuring it out from the lines quoted ...

labman/db/tests/test_process.py
AmandaBirmingham: > Basically here we are generating obs_primer_layout within the test vs. in a single function that generates this result

No, not quite! The big list comprehension here is *not* generating the layout, just *reading* the layout generated by `LibraryPrepShotgunProcess.create` into a convenience format.

When `LibraryPrepShotgunProcess.create` makes `obs`, it fills the`obs.plates[0].layout` property with a list of lists of lists of Wells.   This makes it very hard to create a test case: you have to know the database id of each of the wells to instantiate all the expected Well objects, and it is prohibitively time-consuming for the code to instantiate them in the test case and compare them to all the Well objects in`obs.plates[0].layout`. 

Because of this, I chose to set up the part of the test that checks whether the correct i5 and i7 primer has been assigned to the correct well using a known-good as a list of lists of lists of *strings*: the human-readable well id--A1, etc, the i7 primer name, and the i5 primer name.  The known-good, in this format, is stored in `SHOTGUN_PRIMER_LAYOUT`.  The list comprehension you see is simply **extracting** those strings from`obs.plates[0].layout` (into a convenience variable `obs_primer_layout`) so that they can be compared with the known-good via `assertListEqual`.

In the cases where a given Well object in `obs.plates[0].layout` is None, of course you can't access its various nested string properties, so None is returned instead of the strings.

If you have thoughts for how to adjust the list comprehension to make it more readable, that would be great!

labman/gui/handlers/process_handlers/test/test_gdna_compression_process.py
AmandaBirmingham: The original test was "compressing" the SAME plate (plate 22) twice onto a larger plate.  According to Greg, this is not possible, which makes it a bad choice for a test case. Now that we have expanded the test database to actually *have* more than one gdna plate, the test has been changed to reflect the realistic situation of compressing two different gdna plates onto a larger plate rather than the impossible situation of compressing the same plate twice.

labman/gui/templates/normalization.html
AmandaBirmingham: Yes, we have an issue for this already; it is #323 'Fix "reformat" option in Normalize gDNA Plates bug'.

labman/db/process.py
```diff
@@ -697,17 +782,32 @@ def create(cls, user, plates, plate_ext_id, robot):
...
+            get_interleaved_quarters_position_generator(
+                len(plates), plate_config_384.num_rows,
+                plate_config_384.num_columns)
+
+        for p in position_generator:  # each position on interleaved plate
+            input_layout = plate_layouts[p.input_plate_order_index]
+            input_well = input_layout[p.input_row_index][p.input_col_index]
+            # completely empty wells (represented as Nones) are ignored
+            if input_well is not None:
+                # note adding 1 to the row/col from p, as those are
```
antgonza: When I said that it sounded messy is that in this block we +1 and in the block below (another method) we -1 so it sounds like the dev needs to be really aware of when +/-1 vs. accounting for that somewhere else, automatically.

labman/db/tests/test_process.py
antgonza: perhaps worth adding this explanation?

labman/gui/handlers/process_handlers/test/test_gdna_compression_process.py
antgonza: thanks

labman/db/process.py
```diff
@@ -685,6 +767,9 @@ def create(cls, user, plates, plate_ext_id, robot):
             raise ValueError(
                 'Cannot compress %s gDNA plates. Please provide 1 to 4 '
                 'gDNA plates' % len(plates))
+
+        volume = 1  # TODO: where did this magic # come from and can it change?
```
antgonza: Do we have an issue to figure out if `volume = 1` is correct?

labman/db/process.py
AmandaBirmingham: Correct!  num_quarters=1, total_num_rows=2, total_num_cols=4 would return two positions (1/4 of 2x4) and num_quarters=1, total_num_rows=2, total_num_cols=2 would return just one position (1/4 of 2x2).

labman/db/process.py
AmandaBirmingham: Yep: In that case, as in the case of plate.layout mentioned above, getting layout repeatedly in the loop caused test failures and thus required immediate optimization :)

labman/db/tests/test_process.py
AmandaBirmingham: Good idea--will do!

labman/db/process.py
AmandaBirmingham: Didn't before, but now we do :)  ... just added https://github.com/jdereus/labman/issues/341

labman/db/tests/test_process.py
```diff
@@ -988,12 +1038,39 @@ def test_create(self):
...
+        # human-readable well id--A1, etc, the i7 primer name, and the
+        # i5 primer name. The known-good, in this format, is stored in
+        # SHOTGUN_PRIMER_LAYOUT. The list comprehension below is simply
+        # extracting those strings from obs.plates[0].layout (into a
+        # convenience variable obs_primer_layout) so that they can be compared
+        # with the known-good via assertListEqual.
+        # In the cases where a given Well object in obs.plates[0].layout is
+        # None, of course you can't access its various nested string
+        # properties, so None is returned instead of the strings.
+        obs_primer_layout = [[[well.well_id if well is not None else None,
```
wasade: What about rewriting the comprehension to  following? I think it's translated correctly. I've always found nested comprehensions to be hard to grok. But, this is just personal preference, so please do not consider this a blocking comment.

```python
obs_primer_layout = []
for row in obs.plates[0].layout:
    row_detail = []
    for well in row:
        well_detail = []
        well_detail.append(well.well_id if well is not None else None)
        well_detail.append(well.composition.\
                               i7_composition.\
                               primer_set_composition.\
                               external_id)
        well_detail.append(if well is not None else None)
        well_detail.append(well.composition.\
                           i5_composition.\
                           primer_set_composition.\
                           external_id)
        well_detail.append(if well is not None else None)
        row_detail.append(well_detail)
    obs_primer_layout.append(row_detail)```

File: labman/db/tests/test_process.py [hunk 10]
```diff
@@ -1694,6 +1771,2312 @@ def test_generate_prep_information(self):
         exp = {Study(1): SHOTGUN_EXAMPLE}
         self.assertEqual(obs[Study(1)], exp[Study(1)])
 
+
+INTERLEAVED_POSITIONS = [
```
wasade: This list is magical, would it be possible to get a comment to for some context? The ordering of the indices seems to map back to the docstring in `get_interleaved_quarters_position_generator`, but I may be wrong on that. If the ordering does correspond to that docstring, than perhaps just a comment to the effect of "See the docstring for get_interleaved_quarters_position_generator for more information" may be sufficient.

labman/db/tests/test_process.py
```diff
@@ -1694,6 +1771,2312 @@ def test_generate_prep_information(self):
...
+                                                    input_col_index=9),
+    GDNAPlateCompressionProcess.InterleavedPosition(output_row_index=12,
+                                                    output_col_index=19,
+                                                    input_plate_order_index=1,
+                                                    input_row_index=6,
+                                                    input_col_index=9),
+    GDNAPlateCompressionProcess.InterleavedPosition(output_row_index=14,
+                                                    output_col_index=19,
+                                                    input_plate_order_index=1,
+
```
wasade: I'm assuming this is hard coded from the wetlab?

AmandaBirmingham: Yes; I've added an explanatory comment.

labman/db/tests/test_process.py
AmandaBirmingham: I've added an explanatory comment.

labman/db/tests/test_process.py
AmandaBirmingham: That's a good point--it is only for historical reasons that I did this with a list comprehension, and it doesn't have to stay as one.  I've rewritten it as loops, as you suggested, and I hope it is much clearer; I've left the explanatory comment, though, just in case :)

AmandaBirmingham:
@antgonza is it ok with you if I merge this PR, or do you have other requests?

labman/db/process.py
antgonza: @AmandaBirmingham, can you open an issue about this? I'm not sure in which situations we need to +1 and -1 so perhaps you could add an explanation about this.

AmandaBirmingham:
Hi, @antgonza I have put in explanatory comments whenever the +1 or -1 is used, e.g.:

`# Note subtracting 1 from positions in Well, as those`
`# are 1-based while the positions in layout are 0-based.`

and 

`# Note adding 1 to the row/col from p, as those are`
`# 0-based while the positions in Well are 1-based`

Can you clarify for me what you want to me to open an issue about? (Or you could open it, if that is easier than explaining it to me!)  I can't realistically make the indexes of the layout property 1-based, since it is a list of lists ... do you want to change the row and column positions in the database to be 0-based instead of 1-based so they match the layout indexes?  Or do you have something else in mind?

antgonza:
Thanks for the explanation. 

I guess the best will be to make them 0 based so devs don't need to know when they have to +1 / -1 the indices ... I just feel like if someone else needs to make changes to those indexes, they are gonna need to be aware of when +/-1 ...

PR #348, opened 2018-09-05
Use tube identifiers to plate samples

ElDeveloper:
Note, this PR is currently incomplete and still needs a fair amount of work
(mainly surrounding the tests). I've opened this PR to run the full test suite.

### DIFF
--- a/labman/db/composition.py
+++ b/labman/db/composition.py
@@ -541,6 +541,9 @@ class SampleComposition(Composition):
     def specimen_id(self):
         """The specimen identifier
 
+        This identifier is used to associate the labels written on a physical
+        sample to the unique identifiers in the database (i.e. `sample_name`).
+
         Depending on the composition type, this can be one of three things:
         1) If the composition does not belong to a study (blank or unknown
         sample), then content's id will be returned.
@@ -549,7 +552,7 @@ class SampleComposition(Composition):
         3) If the composition belongs to a study and a specimen id column is
         not set for this study, then the sample id is returned.
         """
-        # figure out if it is a sample or something else
+        # figure out if it is a sample or blank/unknown sample/control
         study = self.study
         if study is not None:
             return study.sample_id_to_specimen_id(self.sample_id)
--- a/labman/db/composition.py
+++ b/labman/db/composition.py
@@ -539,15 +539,22 @@ class SampleComposition(Composition):
 
     @property
     def specimen_id(self):
-        # could be cut into one single query
-        content = self.content
-        sample_id = self.sample_id
+        """The specimen identifier
 
-        # figure out if it is a sample or a blank
-        if sample_id is not None:
-            return self.study.sample_id_to_specimen_id(sample_id)
+        Depending on the composition type, this can be one of three things:
+        1) If the composition does not belong to a study (blank or unknown
+        sample), then content's id will be returned.
+        2) If the composition belongs to a study and a specimen id column is
+        set for this study, then the specimen id is returned.
+        3) If the composition belongs to a study and a specimen id column is
+        not set for this study, then the sample id is returned.
+        """
+        # figure out if it is a sample or something else
+        study = self.study
+        if study is not None:
+            return study.sample_id_to_specimen_id(self.sample_id)
         else:
-            return content
+            return self.content
 
     @property
     def study(self):
--- a/labman/db/composition.py
+++ b/labman/db/composition.py
@@ -541,10 +541,11 @@ class SampleComposition(Composition):
     def specimen_id(self):
         # could be cut into one single query
         content = self.content
+        sample_id = self.sample_id
 
         # figure out if it is a sample or a blank
-        if self.sample_id is not None:
-            return self.study.sample_id_to_specimen_id(content)
+        if sample_id is not None:
+            return self.study.sample_id_to_specimen_id(sample_id)
         else:
             return content
 
--- a/labman/db/composition.py
+++ b/labman/db/composition.py
@@ -538,6 +538,17 @@ class SampleComposition(Composition):
         return self._get_attr('content')
 
     @property
+    def specimen_id(self):
+        # could be cut into one single query
+        content = self.content
+
+        # figure out if it is a sample or a blank
+        if self.sample_id is not None:
+            return self.study.sample_id_to_specimen_id(content)
+        else:
+            return content
+
+    @property
     def study(self):
         """The study the composition sample belongs to"""
         with sql_connection.TRN as TRN:
--- a/labman/db/study.py
+++ b/labman/db/study.py
@@ -156,68 +156,6 @@ class Study(base.LabmanObject):
 
             return res.pop()
 
-    def _list_sample_ids(self, term, limit, specimen):
-        with sql_connection.TRN as TRN:
-            sql = """SELECT sample_id
-                     FROM qiita.study_sample
-                     WHERE study_id = %s
-                     ORDER BY sample_id
-                     LIMIT %s"""
-            sql_args = [self.id]
-            TRN.add(sql, sql_args)
-            return TRN.execute_fetchflatten()
-
-    def _list_specimen_ids(self, term, limit, specimen):
-        with sql_connection.TRN as TRN:
-            sql = """SELECT {0}
-                     FROM qiita.sample_{1}
-                     ORDER BY {0}
-                     LIMIT %s""".format(specimen, self._id)
-            TRN.add(sql, [limit])
-            return TRN.execute_fetchflatten()
-
-    def _search_by_term_in_sample_id(self, term, limit, specimen):
-        with sql_connection.TRN as TRN:
-            sql = """SELECT sample_id
-                     FROM qiita.study_sample
-                     WHERE study_id = %s AND LOWER(sample_id) LIKE %%%s%%
-                     ORDER BY sample_id"""
-            sql_args = [self.id, term]
-            TRN.add(sql, sql_args)
-            return TRN.execute_fetchflatten()
-
-    def _search_by_term_in_specimen_id(self, term, limit, specimen):
-        with sql_connection.TRN as TRN:
-            sql = """SELECT {0}
-                     FROM qiita.sample_{1}
-                     WHERE LOWER({0}) LIKE %%%s%%
-                     ORDER BY {0}
-                     """.format(specimen, self._id)
-            TRN.add(sql, [term])
-            return TRN.execute_fetchflatten()
-
-    def _search_by_term_in_sample_id_and_limit(self, term, limit, specimen):
-        with sql_connection.TRN as TRN:
-            sql = """SELECT sample_id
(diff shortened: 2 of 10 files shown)


labman/gui/static/js/plateViewer.js
```diff
@@ -451,7 +467,6 @@ PlateViewer.prototype.updateDuplicates = function () {
       var row = elem[0] - 1;
       var col = elem[1] - 1;
       that.wellClasses[row][col].push('well-duplicated');
-      that.data[row][col] = elem[2];
```
ElDeveloper: As far as I can tell, this update is not ncessary since this check is should only update the "styling" of the cells, not the content.

ElDeveloper:
This is now ready for review. I would appreciate a careful review, specifically of the sample plating interface since this is the main aspect that was changed. The main change is that when a study has a "specimen_id_column" set, then the user can plate using the values in that column. When this column is not set, the user has to plate the samples using the sample identifiers as defined by Qiita. See this example for the same study except one has a specimen_id_column and one doesn't:

## Without specimen_id_column (users need to type the qiita id):
screen shot 2018-09-06 at 5 43 37 pm

## With a specimen_id_column (users need to type the specimen identifiers):
screen shot 2018-09-06 at 5 33 16 pm


Here's a few notes:
- I've updated the Travis build to use the current dev branch of Qiita. Since the master branch didn't have the most recent changes that are needed for this PR. This can be changed at a later time.
- The PATCH request that dealt with the well updates will now handle using the specimen identifiers if such a column is present in the study's description.
- Since I cannot modify the specimen_id_column attribute via the Python API (because labman doesn't have access to that attribute), I've decided to instead make direct queries against the DB.
- Should we allow users to plate samples even if no study has been selected? I am not sure why this is possible, but it is. I would be happy to remove this altogether if we can all agree on that.

labman/db/study.py
```diff
@@ -64,6 +64,120 @@ def creator(self):
...
+
+        Returns
+        -------
+        str
+            The specimen identifier for this sample identifier.
+
+        Raises
+        ------
+        ValueError
+            If not matches are found (when a specimen_id_column is set).
```
wasade: not -> no

labman/db/study.py
```diff
@@ -64,6 +64,120 @@ def creator(self):
...
+
+        Notes
+        -----
+        If a specimen identifier column hasn't been set, this method will
+        return the inputed value.
```
wasade: inputed -> input

labman/db/tests/test_study.py
```diff
@@ -82,6 +83,107 @@ def test_samples(self):
...
+                           'SKM4', 'SKM5', 'SKM6',
+                           'SKM7', 'SKM8', 'SKM9']
+            self.assertEqual(s.samples('SKM'), exp_samples)
+            exp_samples = ['SKM1', 'SKM2', 'SKM3',
+                           'SKM4', 'SKM5', 'SKM6',
+                           'SKM7', 'SKM8', 'SKM9']
+            # case insensitive
+            self.assertEqual(s.samples('Skm'), exp_samples)
+            self.assertEqual(s.samples('64'), [])
+
```
wasade: Is there a test of a `limit` but no `term`?

labman/db/tests/test_study.py
```diff
@@ -82,6 +83,107 @@ def test_samples(self):
...
+            obs = s.sample_id_to_specimen_id('1.SKM4.640180')
+            self.assertEqual(obs, 'SKM4')
+
+            obs = s.specimen_id_to_sample_id('SKM4')
+            self.assertEqual(obs, '1.SKM4.640180')
+
+            # should be an exact match
+            with self.assertRaisesRegex(ValueError,
+                                        'Could not find \"1\.skm4\.640180\"'):
```
wasade: Just to check, is case insensitivity a safe property of `sample_name`?

labman/db/composition.py
```diff
@@ -537,6 +537,25 @@ def content(self):
...
+        set for this study, then the specimen id is returned.
+        3) If the composition belongs to a study and a specimen id column is
+        not set for this study, then the sample id is returned.
+        """
+        # figure out if it is a sample or something else
+        study = self.study
+        if study is not None:
+            return study.sample_id_to_specimen_id(self.sample_id)
+        else:
+            return self.content
```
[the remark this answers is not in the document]
ElDeveloper: Yes

labman/db/tests/test_study.py
ElDeveloper: Yes there's a test in line 111.

labman/db/tests/test_study.py
ElDeveloper: Yes I think its fair to require exact case matching.

labman/db/tests/test_study.py
wasade: oh, sorry!

ElDeveloper: 👍

labman/db/study.py
```diff
@@ -77,25 +191,31 @@ def samples(self, term=None, limit=None):
...
-                sql += " LIMIT %s"
-                sql_args.append(limit)
+        # an empty wildcard will match all samples i.e. when term is None
+        term = '%%' if term is None else '%%%s%%' % term.lower()
 
+        with sql_connection.TRN as TRN:
+            if specimen_id_column is None:
+                sql = """SELECT sample_id
+                         FROM qiita.study_sample
+                         WHERE study_id = %s AND LOWER(sample_id) LIKE %s
```
AmandaBirmingham: Why is it preferable to get sample_id from qiita.study_sample rather than getting it from the qiita.sample_{studyid} table as is done if there is a specimen_id_column?  Is it better performing or something?  Just curious because I would think that it would be possible (although maybe not advisable? :) to use the current SQL for when there is specimen id in BOTH these cases by doing the below before formatting the sql ...

     if specimen_id_column is None:
            specimen_id_column = "sample_id"

labman/db/tests/test_composition.py
```diff
@@ -115,6 +116,28 @@ def test_sample_composition_get_control_samples(self):
         self.assertEqual(SampleComposition.get_control_samples('TrOL'),
                          ['vibrio.positive.control'])
 
+    def test_sample_composition_specimen_id(self):
+        obs = SampleComposition(1).specimen_id
+        # returns the underlying id if no specimen_id_column is set
+        self.assertEqual(obs, '1.SKB1.640202')
+
```
AmandaBirmingham: Might be nice to have the sort of blank test that is done after setting a specimen_id (copied below) also done here where one isn't set, to make sure we don't do anything later to screw up the handling of that case :)

            obs = SampleComposition(8).specimen_id
            self.assertEqual(obs, 'blank.21.H1')

labman/gui/handlers/process_handlers/test/test_sample_plating_process.py
```diff
@@ -100,12 +108,77 @@ def test_sample_plating_process_handler_patch_request(self):
...
+        # HACK: the Study object in labman can't modify specimen_id_column
+        # hence we do this directly in SQL, if a test fails the transaction
+        # will rollback, otherwise we reset the column to NULL.
+        sql = """UPDATE qiita.study
+                 SET specimen_id_column = %s
+                 WHERE study_id = 1"""
+        with sql_connection.TRN as TRN:
+            TRN.add(sql, ['anonymized_name'])
+
+            print(Study(1).specimen_id_column)
```
AmandaBirmingham: What is the print for?

labman/gui/handlers/process_handlers/test/test_sample_plating_process.py
```diff
@@ -100,12 +108,77 @@ def test_sample_plating_process_handler_patch_request(self):
...
-            user, 10, 'replace', '/well/8/1/notes', 'New Notes', None)
+            self.user, 10, 'replace', '/well/8/1/1/notes', 'New Notes', None)
         self.assertEqual(tester.notes, 'New Notes')
         obs = sample_plating_process_handler_patch_request(
-            user, 10, 'replace', '/well/8/1/notes', '  ', None)
+            self.user, 10, 'replace', '/well/8/1/1/notes', '  ', None)
         self.assertIsNone(tester.notes)
 
+    def test_patch_with_specimen_id(self):
+        # Test success
```
AmandaBirmingham: I'm confused by this comment ... the tests right below this comment are just showing the well is what we expect it to be when we start the test, right?  Maybe this comment could be expanded a little bit to make it more obvious?

labman/gui/static/js/plateViewer.js
```diff
@@ -352,6 +352,20 @@ PlateViewer.prototype.loadPlateLayout = function () {
...
 };
 
+/**
+ * Fetch the study that is currently selected in the UI
+ */
+PlateViewer.prototype.getActiveStudy = function () {
+  studyID = get_active_studies().pop();
+
+  if (studyID === undefined) {
+    alert('Please select a study, plating samples without selecting a study ' +
```
AmandaBirmingham: study, -> study;

ElDeveloper:
Thanks for the review @AmandaBirmingham!

> I don't think I understand how the functionality works when samples from more than one study are being plated on the same plate. How does the user switch between studies to pick their preferred active one?

This functionality was already provided with the current interface, so it didn't require any additional changes. For example see in this animated gif how I have two studies to plate both of which have sample names with the letter `s`. However, depending on the study that is selected the suggestions and validations will pop up as expected. At the end of the animation, I add and select another study and type in the letter `a` which reveals only sample names for that third study.

study-selection

Let me know if you would rather go through this in person, I can come find you by your office. In the meantime I'll finish addressing your review comments.

labman/db/study.py
ElDeveloper: This is a great suggestion!

labman/gui/handlers/process_handlers/test/test_sample_plating_process.py
ElDeveloper: My bad! That has been removed now.

labman/db/study.py
```diff
@@ -64,6 +64,120 @@ def creator(self):
...
+                         """.format(self._id, specimen_id_column)
+            TRN.add(sql, [specimen])
+            res = TRN.execute_fetchflatten()
+
+            if len(res) == 0:
+                raise ValueError('Could not find "%s"' % specimen)
+            # if a specimen_id_column is not unique (since this is softly
+            # enforced), then there can be more than one match
+            elif len(res) > 1:
+                raise RuntimeError('More than one match was found, there is a '
```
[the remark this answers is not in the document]
ElDeveloper: Good suggestion.

labman/db/tests/test_composition.py
ElDeveloper: Great suggestion.

wasade:
Awesome!

AmandaBirmingham:
Agreed!  Thanks, @ElDeveloper!

Issue #351, opened 2018-09-19
Centralize addPlate functionality in front-end html/javascript


AmandaBirmingham:
Currently, the following pages all allow users to select plates from a list and then add the selected plates as divs on the page:

normalization.html
compression.html
extraction.html
library_pooling.html
library_prep_shotgun.html
library_prep_16S.html
parse_quantification.html

Unfortunately, all of these pages have their own separate code to do this.  A conceptual bug was recently found in compression.html that, when prepopulating the gui with plates selected earlier, does so out of order (see #325 ).  The same issue will affect at least normalization.html, library_pooling.html, and library_prep_shotgun.html because they also sometimes prepopulate info on multiple plates; however because the order in which plates are shown on these pages is not itself conveying information (unlike on the compression.html page, where it indicates which gdna plate went into which quadrant of the compression plate!) this issue does not HAVE to be fixed on these pages.  Still, someday it would be nice to replace the copy-and-pasted code in these pages with the new, centralized code use to fix compression.html (which is in the new plateList.js file) to address this little niggle and to make it easier to centrally fix any future bugs that are found in the plate-addition functionality.

AmandaBirmingham:
Closed by PR #468

PR #352, opened 2018-09-19
fixes #325 Enforce correct order of plates in "Compress gDNA plates" interface

### LINKED ISSUE #325: Enforce correct order of plates in "Compress gDNA plates" interface
If one goes to the plate list and clicks on the eye icon for a compressed gDNA plate, one is taken to the "Compress gDNA plates" interface, which displays the gDNA plates that were aggregated onto that compressed gDNA plate.  The order in which the plates are displayed on the screen is the ONLY way for the user to find out, from the Labman GUI, which plates were placed where on the compressed plate.   The order MUST BE as follows:

1) the top-left-most plate (original wells transferred into A1, C1, E1, G1, I1, K1, M1, O1, A3, C3, etc)
2) the top-right-most plate (original wells transferred into A2, C2, E2, G2, I2, K2, M2, O2, A4, C4, etc)
3) the bottom-left-most plate (original wells transferred into B1, D1, F1, H1, J1, L1, N1, P1, B3, D3, etc)
4) the bottom-right-most plate (original wells transferred into B2, D2, F2, H2, J2, L2, N2, P2, B4, D4, etc)

However, currently the interface displays the plates in random order that sometimes differs between page accesses for the same plate, as shown by these two screenshots of the interface for the same test plate, taken 3 minutes apart:

screen shot 2018-08-14 at 5 37 37 pm

screen shot 2018-08-14 at 5 40 20 pm

AmandaBirmingham:
See #325 for a description of the issue and why it is a must-fix on this page.  See #351 for why it ISN'T a must-fix on the three other pages it affects, although it would certainly be nice to centralize the copy-pasted code ... someday :)

### DIFF
--- a/labman/gui/static/js/plateList.js
+++ b/labman/gui/static/js/plateList.js
@@ -1,0 +1,44 @@
+var gPlateInfoByPlateId = {};
+
+function getPlateInfoRequest(plateId, postCheckFunc){
+    var request = $.get('/plate/' + plateId + '/', function (data) {
+        gPlateInfoByPlateId[plateId] = data;
+    })
+      .fail(function (jqXHR, textStatus, errorThrown) {
+        bootstrapAlert(jqXHR.responseText, 'danger');
+        $('#addPlateModal').modal('hide');
+      });
+
+    if (typeof(postCheckFunc)!=='undefined'){
+        postCheckFunc()
+    }
+
+    return request;
+}
+
+function requestAndAddPlates(relevantPlateIds, postRequestFunc, postAddFunc){
+    // do asynchronous get requests to get info for all the relevant plates
+    var requests = [];
+    $.each(relevantPlateIds, function(idx, plateId) {
+      requests.push(getPlateInfoRequest(plateId, toggleAddPlate));
+    });
+
+    // below is basically, "do this when an unknown number of requests have all finished".
+    // "$.when takes any number of parameters and resolves when all of these have resolved";
+    // see https://stackoverflow.com/a/14777262 for details on construct
+    $.when.apply($, requests).then(function() {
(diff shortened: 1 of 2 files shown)


wasade:
I didn't see anything on review, however I'm not very proficient with JS. I was able to understand the motivations and logic flow, and the contextual comments (e.g., `addPlate`) were very helpful.

labman/gui/static/js/plateList.js
```diff
@@ -0,0 +1,44 @@
+var gPlateInfoByPlateId = {};
```
ElDeveloper: Does this variable need to be cleared somewhere else in the UI or is an initialization per-page-load sufficient to guarantee correct functionality.

AmandaBirmingham: This is a good question.  I believe that initialization per-page-load is sufficient, particularly because this variable is being accessed by specific keys; there is no time when the code loops over ALL keys in the dictionary without regard for whether they are also in the explicit list of plates to display, so even if there were some records left over for e.g. removed plates, they should not be accessed or acted upon.  

Now, that said, it is always possible someone down the line could change how this variable is being accessed.  I'm open to suggestions if there's a variable-management approach you think would be preferable!

ElDeveloper: I think this should be fine then. In order to get rid of this global variable I think we would need to refactor the way the add plate functionality is working so that everything was in the same scope. Perhaps my only suggestion would be to add a comment in line 71 of `labman/gui/templates/compression.html` noting where this variable is being declared/populated.

AmandaBirmingham: Sold :)

PR #360, opened 2018-10-03
Fix breaking issues identified during initial real-data test

### LINKED ISSUE #353: Prevent unhandled exception when trying to plate a sample without a study selected
Trying to enter a sample into a plate without picking a study causes a back-end exception.
Steps to reproduce:
1) Create a new sample plate and do NOT pick a study to go with it (or open an existing sample plate that has no study associated with it).
2) Type or paste in any sample name into at least one cell.
3) Receive a client-side alert saying plating without a sample is not allowed (see image).  If you are pasting in multiple sample ids, you will receive one alert for each sample id.
screen shot 2018-10-01 at 4 34 16 pm
4) After clicking "OK" on the client-side alert (multiple times if entering multiple sample ids) , receive a server-side exception message (see image).
screen shot 2018-10-01 at 4 34 28 pm

### LINKED ISSUE #356: Reduce number of GET calls during sample plating
The modifyWell function in plateViewer.js calls two functions every time a well is successfully modified:

           that.updateDuplicates();
           that.updateUnknown();

*Each* of these functions does a `$.get('/plate/' + this.plateId + '/')` and then updates some display info on the page based on the returned data; however, the returned data is not changed by the GET call, so *they both get the exact same data*, one directly after another.  This means that every well change generates 2 GET calls, whereas only one is needed.  While this is not precisely a bug, it is contributing to time-out errors when pasting in whole plate layouts at once, since 192 GET calls (in the case of a 96-well plate) are being made.

### LINKED ISSUE #357: Prevent uninformed discarding of plates
Clicking the red X on a plate in the plate list (see below) instantly marks it as "discarded" in the database, with no warning given:
screen shot 2018-10-02 at 3 53 18 pm

After discarding point it no longer appears in the plate list (see below example for abtest_get5) but it is not actually deleted, so its name cannot be reused, and there is no way to short of going manually into the database to resurrect it:
screen shot 2018-10-02 at 3 53 31 pm

### LINKED ISSUE #358: Include database permission requirements in install directions
The current install directions will not produce a working labman install unless the supplied database user is granted permissions to the labman database.  Incorporate the info @jdereus  provided about how to do this into the install document:

Create user labman
Alter user labman with encrypted password ‘<password here>’;
 
Grant select on all tables in schema qiita to labman;
 
As labman should not have create permissions on db, the labman patch scripts will be run as postgres.  This gives odd ownerships that need to be modified.
 
-bash-4.2$ SCHEMA=labman
-bash-4.2$ DB=qiita
-bash-4.2$ for table in `psql -tc "select tablename from pg_tables where schemaname = '${SCHEMA}';" ${DB}` ; do psql -c "alter table ${SCHEMA}.${table} owner to labman" ${DB}; done
 
Grant all on all tables in schema labman to labman;

Also, point out that when starting the labman server, if it should be started on any port other than the default port of 8080, the user needs to run start_webserver specifying the optional `--port` switch, e.g. `--port 5555`.

### LINKED ISSUE #359: Upgrade command-line command names to work with click library version 7
The labman.py and db/tests/tester.py scripts use the click library to create command-line command names.  The click library has just released a new major version, which now automatically changes command names with underscores to have dashes instead unless you specifically set the name, and this causes the travis build script to fail because some of the commands are not found:

screen shot 2018-10-03 at 11 27 48 am

I guess this is a bug ... now ... https://github.com/pallets/click/issues/1123

AmandaBirmingham:
Closes #356 Reduce number of GET calls during sample plating
Closes #353 Prevent unhandled exception when trying to plate a sample without a study selected
Closes #359 Upgrade command-line command names to work with click library version 7
Closes #357 Prevent uninformed discarding of plates
Closes #358 Include database permission requirements in install directions

### DIFF
--- a/labman/gui/handlers/process_handlers/sample_plating_process.py
+++ b/labman/gui/handlers/process_handlers/sample_plating_process.py
@@ -77,6 +77,14 @@ def sample_plating_process_handler_patch_request(
             study_id = int(req_path[3])
             well_attribute = req_path[4]
 
+            # The default values of the variables sample_id and
+            # blank_or_unknown are set before the try, and these are the values
+            # that are used if either (a) study_id is not 0 OR (b) the try
+            # fails with a ValueError (see comment in try below). If the try
+            # fails with anything other than a ValueError then the entire
+            # function bails, so it doesn't matter what these variables are set
+            # to. If and only if the try succeeds are these variables set to
+            # the values within the try.
             sample_id = req_value
             blank_or_unknown = True
 
@@ -85,6 +93,16 @@ def sample_plating_process_handler_patch_request(
             # the fully qualified sample id(s)--i.e., <studyid>.<sampleid>.
             if study_id != 0:
                 try:
+                    # Note that the try fails iff the
+                    # Study(study_id).specimen_id_to_sample_id() call fails,
+                    # as the blank_or_unknown = False can't really fail ...
+                    # this assures that we can't realistically end up in an
+                    # inconsistent situation where sample_id has been set
+                    # during the try but blank_or_unknown has not.
+                    #
+                    # Thus, the ordering of these two statements within the try
+                    # is really important: do not change it unless you
+                    # understand all of the above!
                     sample_id = Study(study_id).specimen_id_to_sample_id(
                         req_value)
                     blank_or_unknown = False
--- a/labman/gui/handlers/process_handlers/sample_plating_process.py
+++ b/labman/gui/handlers/process_handlers/sample_plating_process.py
@@ -77,12 +77,21 @@ def sample_plating_process_handler_patch_request(
             study_id = int(req_path[3])
             well_attribute = req_path[4]
 
-            blank_or_unknown = False
-            try:
-                sample_id = Study(study_id).specimen_id_to_sample_id(req_value)
-            except ValueError:
-                sample_id = req_value
-                blank_or_unknown = True
+            sample_id = req_value
+            blank_or_unknown = True
+
+            # It actually IS possible to plate a plate without specifying
+            # separate study id(s); can plate just all blanks, or can provide
+            # the fully qualified sample id(s)--i.e., <studyid>.<sampleid>.
+            if study_id != 0:
+                try:
+                    sample_id = Study(study_id).specimen_id_to_sample_id(
+                        req_value)
+                    blank_or_unknown = False
+                except ValueError:
+                    pass
+                # end try/except
+            # end if
 
             if well_attribute == 'sample':
                 if req_value is None or not req_value.strip():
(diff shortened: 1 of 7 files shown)


labman/gui/handlers/process_handlers/sample_plating_process.py
```diff
@@ -77,12 +77,21 @@ def sample_plating_process_handler_patch_request(
...
+            blank_or_unknown = True
+
+            # It actually IS possible to plate a plate without specifying
+            # separate study id(s); can plate just all blanks, or can provide
+            # the fully qualified sample id(s)--i.e., <studyid>.<sampleid>.
+            if study_id != 0:
+                try:
+                    sample_id = Study(study_id).specimen_id_to_sample_id(
+                        req_value)
+                    blank_or_unknown = False
```
wasade: Should the setting of this variable be done within `except`?

AmandaBirmingham: Counterintuitively, no :)  The default values of the variables `sample_id` and `blank_or_unknown` is now done before the `try`, and these are the values that are used if either (a) `study_id` is not 0 OR (b) the `try` fails with a `ValueError` (see * below).  If the `try` fails with anything else then the entire function bails, so it doesn't matter what those variables are.  If and only if the `try` succeeds are the variables set to the values within the `try`.


* Note that the `try` fails iff the `Study(study_id).specimen_id_to_sample_id()` call fails, as the `blank_or_unknown = False` can't really fail ... this assures that we can't realistically end up in an inconsistent situation where `sample_id` has been set during the `try` but `blank_or_unknown` has *not*.  Thus, the ordering of the two statements within the `try` is really important.

wasade: ahhhh, okay, thank you for the explanation!

labman/gui/static/js/plateViewer.js
```diff
@@ -359,12 +359,10 @@ PlateViewer.prototype.getActiveStudy = function () {
   studyID = get_active_studies().pop();
 
   if (studyID === undefined) {
-    alert('Please select a study; plating samples without selecting a study ' +
-          'is disallowed');
-    return;
+    return 0;
```
ElDeveloper: Would be good to document this behavior in the docstring of this method, and state the implication.

labman/gui/static/js/plateViewer.js
```diff
@@ -450,54 +447,49 @@ PlateViewer.prototype.updateAllRows = function () {
   for (var i = 0; i < this.rows; i ++ ) {
     this.grid.updateRow(i);
   }
-}
+};
+
 
-PlateViewer.prototype.updateDuplicates = function () {
+PlateViewer.prototype.updateUnknownsAndDuplicates = function () {
```
ElDeveloper: Good catch on refactoring this. I believe some comments in this file refer to the `updateDuplicates` and the `updateUnknowns` methods. It would be good to update/remove where appropriate.

Issue #369, opened 2018-10-09
Support "externally extracted" gDNA plates in production-like databases


AmandaBirmingham:
Although labman is intended to support externally extracted gDNA plates, that functionality fails when run against a production-like database:

![screen shot 2018-10-09 at 10 30 55 am](https://user-images.githubusercontent.com/10677935/46686927-74418880-cbae-11e8-9b59-ef39a7ee64eb.png)

This is because externally extracted plates depend on a number of dummy records in the database, and those records are added in the `populate_test_db.sql` script, which is not run when a production-like database is created.

PR #376, opened 2018-10-29
Support externally extracted gDNA plates in production

AmandaBirmingham:
TL;DR: Fix for #369 Support "externally extracted" gDNA plates in production-like databases

During real-data testing against labman on qiita-rc, I discovered that the externally extracted gDNA plates functionality did not work because it depend on a number of dummy records in the database, and those records are added in the populate_test_db.sql script--which is not run when a production-like database is created.  This PR moves that functionality out of  populate_test_db.sql 
 and into db_patch_manual.sql--and then updates what-feels-like-zillions of hard-coded process and composition ids in the tests that are affected by having those records created before the test records rather than after them.


ElDeveloper:
Thanks for such a thorough effort @AmandaBirmingham!

PR #378, opened 2018-11-01
Blank, source sample name, and output file renaming

### LINKED ISSUE #374: Modify all output file names to match specifications from the wet lab
Per greg, gail, rodolfo, mackenzie 10/09/2018

shotgun workflow:
normalization echo picklist: YYYY-mm-dd_nameofcompressedplate_input_norm.txt
library prep echo picklist: YYYY-mm-dd_nameofcompressedplate_indices.txt
pool file: YYYY-mm-dd_nameofcompressedplate_normpool.csv
sample sheet: YYYY-mm-dd_samplesheet_runname_runexperiment_knight.csv

amplicon workflow:
pool file: YYYY-mm-dd_nameofplate_normpool.csv
sample sheet: YYYY-mm-dd_samplesheet_runname_knight.csv
prep sheet: YYYY-mm-dd_prep_primertype (16S, ITS, 18S)_runname.txt

### LINKED ISSUE #375: Name blanks with plate name, not plate id
Per greg, gail, rodolfo, mackenzie 10/09/2018

Instead of blanks being named something like "blank.67.G11" where 67 is the id of the plate on which they occur, they should be named something like "blank.salk.ting.II.4.G11" where salk ting II" is the external_id of the plate (with id 67) and that external id has had spaces replaced with periods throughout.

### LINKED ISSUE #377: Use underscores instead of whitespace in Source Plate Names in shotgun library prep echo picklists
Per Greg H. and Rodolfo 10/09/2018

AmandaBirmingham:
Addresses 3 issues logged in 10/09/2018 meeting with wet lab and Gail:

Fixes #374 Modify all output file names to match specifications from the wet lab
Fixes #375 Name blanks with plate name, not plate id
Fixes #377 Use underscores instead of whitespace in Source Plate Names in shotgun library prep echo picklists


labman/db/composition.py
```diff
@@ -478,16 +479,23 @@ def _get_sample_composition_type_id(compostion_type):
             sct_id = TRN.execute_fetchlast()
         return sct_id
 
+    @staticmethod
+    def generate_content(sample_name, well):
```
wasade: Would it be possible to have at least a 1 line docstring?

labman/gui/handlers/process_handlers/normalization_process.py
```diff
@@ -64,16 +64,11 @@ def post(self):
...
         text = process.generate_echo_picklist()
-
-        self.set_header('Content-Type', 'text/csv')
-        self.set_header('Expires', '0')
-        self.set_header('Cache-Control', 'no-cache')
-        self.set_header('Content-Disposition', 'attachment; filename='
-                        'NormalizationSheet_%s.txt' % process_id)
-        self.write(text)
-        self.finish()
+        compressed_plate_name = process.compressed_plate.external_id
```
wasade: This is within the `DownloadNormalizationProcessHandler`, and does not use `process.nornalization_process`. In the prior file, it is `DownloadLibraryPrepShotgunHandler`, and it makes a very similar call but does use `process.normalization_process`. Is this correct?

labman/gui/handlers/process_handlers/pooling_process.py
```diff
@@ -416,24 +415,24 @@ def post(self):
...
-                        '%s' % filename)
-        self.write(text)
-        self.finish()
+        plate_names_of_components = [x[0].container.plate.external_id for x in
+                                     process.components]
+
+        plate_names_set = set(plate_names_of_components)
+        if len(plate_names_set) > 1:
+            raise ValueError("Unable to generate normpool file name for pool"
+                             "based on more than one plate: " +
```
wasade: I think a space is needed before "based"

wasade: Should ´process.components´ raise on this condition or is it valid in other contexts?

labman/gui/handlers/process_handlers/normalization_process.py
AmandaBirmingham: Yes, for the rather esoteric reason that the `process` variable holds an instance of a different subclass of the Process class in each of those two places :)  In this location (`DownloadNormalizationProcessHandler`), `process` IS a `NormalizationProcess` object--which doesn't have a `normalization_process` attribute since, well, it already is one ... we can therefore go straight to looking at the `compressed_plate` attribute to find out on what gdna plate this `NormalizationProcess` was performed. However, in `DownloadLibraryPrepShotgunProcessHandler`, `process` is instead a `LibraryPrepShotgunProcess` object.  A shotgun library prep process is predicated upon an upstream normalization process, so it has a `normalization_process` attribute, and we have to dereference that attribute in order get a to `NormalizationProcess` object that we can interrogate for its  `compressed_plate` attribute.  Does this clarify?

labman/gui/handlers/process_handlers/pooling_process.py
AmandaBirmingham: Good catch re the space before "based"!  

This error definitely shouldn't be raised on process.components directly, because PoolingProcess objects (what `process` is in this context :) definitely *can* have components from multiple plates--that is exactly what we have when we make amplicon sequencing pools (which are pools of plate pools, and thus draw from multiple plates), and in fact a user COULD even validly chose to make a "plate pool" from more than one amplicon library prep plate (see screenshot).  

screen shot 2018-11-01 at 1 58 35 pm

However, the wet lab apparently currently chooses not to do this, and instead chooses to make one pool per library prep plate.  Given this self-imposed limitation, they expect to be able to have the (one-and-only) library plate name embedded in the name of the resulting normpool file.  That *file naming convention* won't work--or at least, won't work as they expect it to--if there are multiple plates represented in the pool, so the error is at the point where the *file name* is generated.  If they decide they want to allow themselves to make plate pools from multiple plates, all they need to do is decide on a more flexible naming convention, and tell us, and we can change the *naming code* and remove this error condition.

I have added the description above into the code as a comment so hopefully it won't be forgotten in the future :)

labman/gui/handlers/process_handlers/normalization_process.py
wasade: Ah okay, thank you for the clarification!

labman/gui/handlers/process_handlers/pooling_process.py
wasade: Thank you as well for this explanation, this makes sense.

labman/gui/handlers/base.py
```diff
@@ -72,3 +74,35 @@ def get(self):
...
+
+    @authenticated
+    def deliver_text(self, name_pieces, process, text, extension="txt"):
+        output_name = self.generate_file_name(name_pieces, process, extension)
+        self._deliver_file(text, output_name, 'text/csv')
+
+    @authenticated
+    def deliver_zip(self, name_pieces, process, text, extension="zip"):
+        output_name = self.generate_file_name(name_pieces, process, extension)
+        self._deliver_file(text, output_name, 'application/zip')
```
ElDeveloper: I'll leave this up to you but I think the deliver_zip method might be more convenient if it can take the content (whatever it is) and compress it. Alternatively, I believe the parameter named `text` should probably be changed as it is (currently) expected to be a zip archive (I think).

labman/gui/handlers/process_handlers/pooling_process.py
```diff
@@ -416,24 +415,39 @@ def post(self):
...
-                   process.pool.container.external_id), process.id)
-
-        self.set_header('Content-Type', 'text/csv')
-        self.set_header('Expires', '0')
-        self.set_header('Cache-Control', 'no-cache')
-        self.set_header('Content-Disposition', 'attachment; filename='
-                        '%s' % filename)
-        self.write(text)
-        self.finish()
+        plate_names_of_components = [x[0].container.plate.external_id for x in
```
ElDeveloper: Minor, but _I think_ this can be a set comprehension i.e. `{x[0] ......}`.

labman/gui/handlers/process_handlers/test/test_base.py
```diff
@@ -0,0 +1,32 @@
...
+# ----------------------------------------------------------------------------
+
+from unittest import main
+
+from labman.gui.testing import TestHandlerBase
+from labman.gui.handlers.base import BaseDownloadHandler
+from labman.db.process import _Process
+
+
+class TestBaseDownloadHandler(TestHandlerBase):
```
ElDeveloper: Would be nice to have tests for the deliver text and deliver zip methods.

labman/db/composition.py
```diff
@@ -478,16 +479,36 @@ def _get_sample_composition_type_id(compostion_type):
...
+        sample_name : str
+            The base name for the sample, such as 1.SKB1.640202 or blank
+        well: labman.db.container.Well
+            The well object in which this sample is placed
+
+        Returns
+        -------
+        A string of the format samplename.plate_name.wellcolumnletrownum
+        """
+        munged_plate_name = re.sub('\s+', '.', well.plate.external_id)
```
ElDeveloper: Unless I'm missing something, this regex can be changed to `[\s,_]+`, and it should make the next line unnecessary (I checked with [pythex](https://pythex.org/?regex=%5B%5Cs%2C_%5D%2B&test_string=this%20is%20a%20_%20test_a_b_c_____d_E&ignorecase=0&multiline=0&dotall=0&verbose=0)).

labman/gui/handlers/base.py
AmandaBirmingham: You are undoubtedly right about the proposed refactor to deliver_zip, but I don't think I can justify the time to do it before soft launch; I have therefore logged it as an enhancement issue so it doesn't get forgotten (#380  Refactor BaseDownloadHandler.deliver_zip to take in uncompressed content and compress it ).  In the meantime, I have changed the name of the parameter from `text` to `archive`.

labman/gui/handlers/process_handlers/test/test_base.py
AmandaBirmingham: I am sad to admit that I spent *three hours* trying to write a working unit test for deliver_text directly and finally gave up ... the issue I encountered is that I can't use the same idiom used in all the other handler tests (that is, `response = self.get('/the/handler/url")` and then check the contents of the `response` object) because the BaseDownloadHandler isn't directly bound to any url (and I thought it seemed like overkill to bind it to one just for testing, but maybe not).  That means I had to fall back to a much more manual sort of testing as described at https://stackoverflow.com/questions/35560415/how-to-write-unit-tests-for-python-tornado-application#35562775 and I just couldn't get the mocking of various objects right ... I kept running up against errors because deep in the code somewhere the handler object was expecting to hold a response object that in fact didn't exist so methods called on it failed, or some such :(  If there are any working examples of a unit test like this somewhere in the Labman/Qiita/etc codebase, I'd be more than happy to copy off one!

ElDeveloper: Sorry to hear that. In all fairness I didn't think about this enough to realize that this was an edge case. I think it is good to leave as is. This is indirectly exercised by other handlers, so at least we can catch things if they do go wrong.

ElDeveloper:
👍

PR #389, opened 2018-11-13
Issue384

### LINKED ISSUE #384: Modify Sample Plate and Sample Well sources on sample sheet
Per Greg 10/09/2018: 
Sample_Plate should actually be name of sample plate--not extracted, not compressed, etc, but
sample_well should come from COMPRESSED plate

AmandaBirmingham:
Fixes #384 Modify Sample Plate and Sample Well sources on sample sheet per Greg's instruction: Sample_Plate column should actually be name of the original sample plate--not extracted, not compressed, etc, but Sample_Well column should be the well of that sample on the  COMPRESSED GDNA plate.


labman/db/process.py
```diff
@@ -2853,7 +2853,7 @@ def _sequencer_i5_index(sequencer, indices):
     @staticmethod
     def _format_sample_sheet_data(sample_ids, i7_name, i7_seq, i5_name, i5_seq,
                                   sample_projs, wells=None, sample_plates=None,
-                                  description=None, lanes=[1], sep=',',
+                                  description=None, lanes=None, sep=',',
```
wasade: That is a great fine, but also concerning given the type of bugs that could be introduced in a long running system. Do you think an issue should be created to determine whether other defaults like this were set?

PR #409, opened 2019-03-28
fix #276

### LINKED ISSUE #276: decide how to create cookie_secret
cookie_secret is the "key" to generate sessions and currently is random, which is a problem if we want to have multiple tornado workers as all of them should have the same one. Thus, I foresee 2 options: use the same one than qiita or separate; this depends on https://github.com/jdereus/labman/issues/275.

Note that the implementation is the same but the difference of the options is the value that goes into that configuration variable.

antgonza:


README.md
```diff
@@ -76,18 +77,18 @@ If creating a development environment for labman, then run:
...
-to set up the database to support running the unit tests.  Alternately, if 
+to set up the database to support running the unit tests.  Alternately, if
 creating a production or production-like environment, run:
 
 ```bash
 psql -d qiita_test -f labman/db/support_files/populate_prod_db.sql
 ```
 
-Note that the postgres user specified for the labman software (`labman` in the config example above) 
+Note that the postgres user specified for the labman software (`labman` in the config example above)
```
wasade: Labman -> LabControl

README.md
```diff
@@ -76,18 +77,18 @@ If creating a development environment for labman, then run:
...
 must be granted "select" permissions on all tables in the "qiita" schema and "all" permissions on
 all tables in the "labman" schema, as well as being the owner of all tables in the "labman" schema.  
 These ownerships and permissions can be granted with SQL like that shown below, after setting the
-USER variable to be the postgres user specified for the labman software and DB to be the 
+USER variable to be the postgres user specified for the labman software and DB to be the
```
wasade: Same here

README.md
```diff
@@ -105,8 +106,8 @@ Labman is now ready to run.  Start the labman server with:
 labman start-webserver
 ```
 
-If it is running successfully, you will see the message `Labman started on port 8080`.  Note that 
-by default, labman starts on port 8080; if you would like to start it on a different port, 
+If it is running successfully, you will see the message `Labman started on port 8080`.  Note that
```
wasade: :(

antgonza:
@wasade, thanks for review and I agree with those tweaks; however, do you think it will be better to open a new issue (with low priority to change labman -> labcontrol everywhere?

wasade:
Progress now would mean less work later?

wasade:
Seems really unhappy in Travis... :(

wasade:
Some of these times are really high: https://travis-ci.org/jdereus/labman/builds/513228812#L3217

wasade:
Tests work locally right? I worry that the CI environment may be too resource constrained.

antgonza:
I'm now able to reproduce the error locally but no idea why is happening ...

wasade:
At the end of the log, it looks like two processes are listening on the same port, is it possible two servers are being started? By recreating locally, do you mean all the timeouts?

wasade:
...also attempting to recreate locally now too

antgonza:
Yes to all and is that nginx is redirecting 8080 to 8383 and labman is starting on 8080 ...

wasade:
Thanks. Was able to recreate one of the same timeouts locally w/o nginx
(once qiita client used 21174 again). Have some ideas to test; about to
take off will fill in once on wifi again.

wasade:
Just issued a PR against your branch which seems to work locally

labman/gui/testing.py
```diff
@@ -66,5 +66,6 @@ def delete(self, url, data=None, headers=None, doseq=True):
         return self._fetch(url, 'DELETE', headers=headers)
 
     def _fetch(self, url, method, data=None, headers=None):
-        return self.fetch(self.get_url(url), self.stop, method=method,
-                          body=data, headers=headers, timeout=20)
+        self.fetch(self.get_url(url), self.stop, method=method,
+                   body=data, headers=headers)
+        return self.wait(timeout=20)
```
wasade: ...I wasn't able to get this to pass locally as the contents of `self.fetch` needed to be returned to the caller

PR #412, opened 2019-03-31
Issue 246

### LINKED ISSUE #246: In plate list interfaces, should also expose date stamp

wasade:
I believe this fixes #246, specifically it implicitly stores the creation timestamp for a given plate, that value is always returned by `Plate.list_plates`, and the UI has been updated to display.

Note the column widths in the display are not well balanced, but in my opinion that is a logically separate issue. I can solve in this PR if requested but it will require a reasonable degree of investigation.

### DIFF
--- a/labman/db/tests/test_plate.py
+++ b/labman/db/tests/test_plate.py
@@ -8,6 +8,7 @@
 
 from unittest import main
 from types import GeneratorType
+import datetime
 
 from labman.db.testing import LabmanTestCase
 from labman.db.plate import PlateConfiguration, Plate
@@ -184,6 +185,23 @@ class TestPlate(LabmanTestCase):
                    'studies': ['Identification of the Microbiomes '
                                'for Cannabis Soils']}])
 
+    def test_plate_list_include_timestamp(self):
+        # ...limit pathological failures by testing within an hour of creation
+        exp = datetime.datetime.now()
+        exp = datetime.datetime(exp.year,
+                                exp.month,
+                                exp.day,
+                                exp.hour)
+
+        for i in Plate.list_plates():
+            obs = i['creation_timestamp']
+            obs = datetime.datetime(obs.year,
+                                    obs.month,
+                                    obs.day,
+                                    obs.hour)
+
+            self.assertEqual(obs, exp)
+
     def test_plate_list_discarded_functionality(self):
         # test case based on the test_list_plates
         obs = Plate.list_plates()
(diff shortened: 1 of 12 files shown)


wasade:
@charles-cowart and I were just talking, and we need to implement a database patching system for this to work. I'll migrate the system from Qiita into this PR.

wasade:
This PR now includes the central elements of Qiita's patching system to allow for both Python and SQL patches if and as necessary for this project.

labman/db/tests/test_plate.py
```diff
@@ -92,9 +93,20 @@ def test_search(self):
             Plate.search(plate_notes='interesting', well_notes='write',
                          query_type='UNION'), [plate22, plate23])
 
+    def strip_out_creation_timestamp(self, plates):
```
charles-cowart: Is this just a hack to make the changes work with existing tests?

labman/gui/handlers/plate.py
```diff
@@ -81,7 +81,9 @@ def get(self):
                       if plate_type is not None else None)
         only_quantified = True if only_quantified == 'true' else False
 
-        rows_list = [[p['plate_id'], p['external_id'],
+        rows_list = [[p['plate_id'],
+                      p['external_id'],
+                      str(p['creation_timestamp']),
```
charles-cowart: Just to confirm, it's more desirable to have p['creation_timestamp'] retain any time-stamping methods and force a str() for perhaps the most used case, rather than convert it to a string value internally?

scripts/labman
```diff
@@ -100,6 +107,73 @@ def config():
...
+
+    Pulls the current patch from the settings table and applies all subsequent
+    patches found in the patches directory.
+    """
+    get_support_file = partial(join, join(dirname(abspath(__file__)), 
+                                          '../labman/db/support_files'))
+    patches_dir = get_support_file('patches')
+
+    # we are going to open and close 2 main transactions; this is a required
+    # change since patch 68.sql where we transition to jsonb for all info
```
charles-cowart: There might be some confusion at some point down the road, if this comment isn't amended to reference Qiita's patch 68.sql.

labman/db/tests/test_plate.py
wasade: Yes, the existing tests are rather ridged.

labman/gui/handlers/plate.py
wasade: Good point

scripts/labman
wasade: Good catch!

labman/db/tests/test_plate.py
charles-cowart: Gotcha. I can make an issue to revisit the tests later.

wasade:
@charles-cowart, are their outstanding items on the review to address?

charles-cowart:
@wasade Nope! Looks good, thanks!

Issue #372, opened 2018-10-09
Limit i5/i7 primer plate selection in shotgun library prep to relevant working primer plates


AmandaBirmingham:
When preparing a shotgun library, the "Select i5 primer plate" and "Select i7 primer plate" dropdowns both display the same list of choices, which includes ALL working primer plates:

<img width="1156" alt="screen shot 2018-10-05 at 5 05 46 pm" src="https://user-images.githubusercontent.com/10677935/46688789-1c595080-cbb3-11e8-84cc-847bd2f88e7d.png">

<img width="1157" alt="screen shot 2018-10-05 at 5 05 58 pm" src="https://user-images.githubusercontent.com/10677935/46688771-0d729e00-cbb3-11e8-9933-d18b0a41ed02.png">

I guess this isn't strictly speaking a bug, as the correct info exists in the list and the user is not prevented from picking it, but it sure makes this task more confusing and error-prone than it should be.

AmandaBirmingham:
Similar issue exists, in reverse, when doing amplicon library prep:
<img width="990" alt="screen shot 2018-10-26 at 3 55 06 pm" src="https://user-images.githubusercontent.com/10677935/49493168-dcf45c80-f80f-11e8-8bfd-f68eca6ef457.png">

wasade:
Fixed in #418

PR #418, opened 2019-04-02
Issue 372

wasade:
Restricts the primer plates to either shotgun or amplicon depending on the page. An example can be found in the screenshot below which can be contrasted with the example in #372.

Screen Shot 2019-04-01 at 3 01 48 PM

### DIFF
--- a/labman/gui/handlers/process_handlers/library_prep_16s_process.py
+++ b/labman/gui/handlers/process_handlers/library_prep_16s_process.py
@@ -53,7 +53,13 @@ class LibraryPrep16SProcessHandler(BaseHandler):
         tools_tm300_8 = Equipment.list_equipment(
             'tm 300 8 channel pipette head')
         tools_tm50_8 = Equipment.list_equipment('tm 50 8 channel pipette head')
-        primer_plates = Plate.list_plates(['primer'])
+
+        primer_plates = []
+        for pp in Plate.list_plates(['primer']):
+            plate = Plate(pp['plate_id'])
+            if plate.process.primer_set.target_name == 'Amplicon':
+                primer_plates.append(pp)
+
         self.render('library_prep_16S.html', plate_ids=plate_ids,
                     robots=robots, tools_tm300_8=tools_tm300_8,
                     tools_tm50_8=tools_tm50_8, primer_plates=primer_plates,
--- a/labman/gui/handlers/process_handlers/library_prep_shotgun_process.py
+++ b/labman/gui/handlers/process_handlers/library_prep_shotgun_process.py
@@ -39,7 +39,12 @@ class LibraryPrepShotgunProcessHandler(BaseHandler):
             i5plate = process.i5_primer_plate.id
             i7plate = process.i7_primer_plate.id
             volume = process.volume
-        primer_plates = Plate.list_plates(['primer'])
+
+        primer_plates = []
+        for pp in Plate.list_plates(['primer']):
+            plate = Plate(pp['plate_id'])
+            if plate.process.primer_set.target_name == 'Shotgun':
+                primer_plates.append(pp)
 
         self.render('library_prep_shotgun.html', plate_ids=plate_ids,
                     primer_plates=primer_plates, process_id=process_id,


labman/gui/handlers/process_handlers/library_prep_16s_process.py
```diff
@@ -53,7 +53,13 @@ def get(self):
         tools_tm300_8 = Equipment.list_equipment(
             'tm 300 8 channel pipette head')
         tools_tm50_8 = Equipment.list_equipment('tm 50 8 channel pipette head')
-        primer_plates = Plate.list_plates(['primer'])
+
```
antgonza: I think you can use plate_types in list_plates?

labman/gui/handlers/process_handlers/library_prep_shotgun_process.py
```diff
@@ -39,7 +39,12 @@ def get(self):
             i5plate = process.i5_primer_plate.id
             i7plate = process.i7_primer_plate.id
             volume = process.volume
-        primer_plates = Plate.list_plates(['primer'])
+
+        primer_plates = []
```
antgonza: Same here but with Shotgun ...

labman/gui/handlers/process_handlers/library_prep_16s_process.py
wasade: Unfortunately doesn't work. That tests against the descriptipn field in the composition table (or composition types? Can't remember off hand), and the entries in there do not provide sufficient granularity to determine whether shotgun or amplicon

antgonza: Then I'm wondering if perhaps worth fixing there vs. looping over all the plates?

wasade:
That's i suspect a pretty reasonable sized data model change. This doesn't
loop over all plates either, just the primer plates

labman/gui/handlers/process_handlers/library_prep_16s_process.py
wasade: (github didn't reply to the right place via email)

That's i suspect a pretty reasonable sized data model change. This doesn't
loop over all plates either, just the primer plates

wasade: The table type restriction ("primer" in this case) filters on this [column](https://github.com/jdereus/labman/blob/master/labman/db/support_files/db_patch.sql#L5). The distinction, as best I can tell, for amplicon or shotgun is based off this [column](https://github.com/jdereus/labman/blob/master/labman/db/support_files/db_patch.sql#L47) in a separate table `primer_set`. The `primer_set` table is not part of the constructed SQL query used by `list_plates`. `list_plates` is already a complex function. It may be possible to increase its complexity. However I'm not familiar enough with the broader data model to know whether this is an edge case in the data model (i.e., this should be in `composition_types`), because it seems weird that this information is "difficult" to get too. 

As best I can tell, this filtering we're doing here is only done in these locations. We can create a separate issue for decomposing where and how this filtering is being performed, but I would be hesitant to include it in the full launch milestone.

antgonza: After checking the code and testing, it looks like plate_types has to be an _array_ and one of [these values](https://github.com/jdereus/labman/blob/master/labman/db/composition.py#L43-L53); something like this:`list_plates(plate_types=['16S library prep'])`

wasade: That is the wrong information and does not solve this problem. 

```python
In [6]: Plate.list_plates(["16S library prep"])
Out[6]:
[{'plate_id': 23, 'external_id': 'Test 16S plate 1'},
 {'plate_id': 29, 'external_id': 'Test 16S plate 2'},
 {'plate_id': 32, 'external_id': 'Test 16S plate 3'},
 {'plate_id': 35, 'external_id': 'Test 16S plate 4'}]
```

antgonza: Got it now, you are searching for all the primer plates that are Shotgun or Amplicon, thanks for explanation. Would it be more convenient to add that option to the list_plates method? So you can call like:
```
Plate.list_plates() # returns everything
Plate.list_plates(plate_types=['primer']) # returns all primers
Plate.list_plates(plate_types=['primer'], target='Shotgun') # returns Shotgun primers?
```
This should also avoid the duplication of the loops ...

PR #440, opened 2019-04-09
EpMotion input file modified to support Windows

charles-cowart:
EpMotion input file will now be in Windows text format, with CRLF
('\r\n') terminating each line, as opposed to LF ('\n') found on MacOSX/*NIX
systems.


charles-cowart:
Not necessarily - I chalked it up to the unit test using a module like CSV, that employed universal newlines functionality. At some point I think it's worth incorporating the CSV or the 3rd party Illumina format modules around to handle issues like this more gracefully.

The Illumina file format, from what I've seen, is just an INI-like format header within an overall CSV format.

PR #442, opened 2019-04-09
fix #364

### LINKED ISSUE #364: Speed up pool list response or add wait message
Observed on maptest.ucsd.edu labman against qiita-rc db: In a database with only 16 pools, the pool list takes more than two seconds to load; while the load is happening, the page (erroneously) indicates that there are no pools:

While load is happening:
screen shot 2018-10-04 at 4 12 18 pm

After load is complete:
screen shot 2018-10-04 at 4 14 02 pm

I expect that this will be confusing and irritating for users.

antgonza:
To fix #364 I decided to go two ways:
1. Add a message in the GUI so it's clear/visible that we are waiting for the list to be displayed.example
2. Speed up the listing via an improved query, this means that I created a new file db.util where we should store all these improvements from now on. To better reproduce the issue I added a few pools (total 94). With these, the original method takes `7.57 s ± 44.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`, while the new added method takes `134 ms ± 1.6 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)`. Note that there is still room for performance improvement but that will take multiple changes; explained as a comment in the method so based on the time it will take to make these changes and the current performance improvement, I decided to not solve and not add an issue. Once we get to 1000s of pools we can address.

### DIFF
--- a/labman/db/util.py
+++ b/labman/db/util.py
@@ -16,6 +16,13 @@ def get_pools_listing():
     Returns
     -------
     list of list
+
+    Notes
+    -----
+    Reproduces the functionality of:
+    [p.id, p.container.external_id, p.is_plate_pool, p.upstream_process.id]
+     for p in PoolComposition.get_pools()]
+    but with direct calls to the database
     """
     with sql_connection.TRN as TRN:
         sql = """SELECT pool_composition_id, external_id,
--- a/labman/db/util.py
+++ b/labman/db/util.py
@@ -1,0 +1,44 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2017-, labman development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# ----------------------------------------------------------------------------
+
+from . import sql_connection
+from . import process
+
+
+def get_pools_listing():
+    """Generates pool listing for GUI with direct DB access for speed
+
+    Returns
+    -------
+    list of list
+    """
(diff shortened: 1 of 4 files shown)


antgonza:
Assigned 3 reviewers but 1-2 should be fine; just not sure about availability.

labman/db/util.py
```diff
@@ -0,0 +1,44 @@
...
+from . import process
+
+
+def get_pools_listing():
+    """Generates pool listing for GUI with direct DB access for speed
+
+    Returns
+    -------
+    list of list
+    """
```
wasade: Possible for a sentence or two of the constituents of the nested list? Looks like `(pool_composition_id, external_id, is_plate_pool, something)`. I don't know what the `is_plate_pool` or `something` are here though :/

antgonza: To be honest, neither do I ... would it suffice to add something like:
`Reproduces the functionality of [p.id, p.container.external_id, p.is_plate_pool, p.upstream_process.id] for p in PoolComposition.get_pools() but with direct calls to the database`?

charles-cowart:
This seems like a strategy we might need to employ in general to get back more performance.

labman/db/util.py
charles-cowart: I'm good with that.

antgonza: k, added ...

charles-cowart:
Will merge once Travis build completes successfully.

PR #446, opened 2019-04-10
Fixes #429

### LINKED ISSUE #429: In the column "orig_name" study ID should not be prepended
Should be name as provided by user - sample_name or tube_id

charles-cowart:
Prep_file output modified to remove study_id prepended to every entry in
the 'Orig_ID' column. Tests have not been updated. EXPERIMENTAL_SAMPLES_PREP_EXAMPLE and the like almost certainly need another version with the study_ids removed from 'Orig_ID' column.


charles-cowart:
Request any two out of the three reviewers to review and merge

charles-cowart:
It's a good point to keep in mind, although Gail has reviewed the sheets already and she hasn't brought this up as an issue. It seems like the final output has headers and output tweaked to be compatible with legacy processes, etc. I see this tweak as another instance of that.

antgonza:
K, thanks. I think this only needs 1 reviewer ... let me know if you want me to merge ...

labman/db/process.py
```diff
@@ -3563,6 +3563,13 @@ def generate_prep_information(self):
...
             df = df[order]
             sio = StringIO()
+
+            # remove Study ID prepended to each entry in column 'Orig_name'
+            # before writing out to disk.
+            df['Orig_name'].replace(regex=True,
+                                    inplace=True,
+                                    to_replace=r'^\d+\.',
+                                    value=r'')
             df.to_csv(sio, sep='\t', index_label='sample_name')
```
AmandaBirmingham: Is there anywhere in qiita that the sample name exists WITHOUT the study id prepended to it?  The orig_name we are working with comes from the following SQL:

`study_sample.sample_id AS orig_name`

It would be much preferable to pull the info we want from the db (if it exists there) rather than use a heuristic to cut it off afterwards!

antgonza: No, it doesn't - we overwrite the sample names. However, I have seen users that add a column with that information, which is basically a duplicated of sample_name/sample_id.

charles-cowart: @AmandaBirmingham I agree in principle, but it is as @antgonza says; I can say though that the study_id.sample_id pattern is very durable, and I confirmed this pattern held through across the unit-test study. I can also confirm it with all rows in study_sample, if you'd like.

AmandaBirmingham: Ok, but we definitely DO have the study id somewhere in the qiita information. We are literally pulling it in the same sql query that gets orig_name:

            `SELECT`
               ` study.study_id, study_sample.sample_id,`

It seems to me that we therefore should not NEED to use a regex (which could break silently if the pattern changes, although I acknowledge that probably it won't) when we could just replace the known study_id (+ a dot) within the sample_id with "" ...

charles-cowart: A fair point. I can put the change in after our meeting. I suppose it's worth mentioning that in either case, the result would be no change if the assumption doesn't hold. I bring it up because my assumption is that Gail or someone would notice those right away, and as we wouldn't know how to handle this case, it's a proper mode of failure.

charles-cowart: @AmandaBirmingham After some additional time and effort spent on this issue, it seems as though a solution using study.study_id is not workable, as there are some cases where study_id returns, and is expected to return NULL (See line 3497). Reverting back to the previous solution, which I believe is suitable.

wasade: If `study_id` is `None`, then that means Qiita does not have the sample metadata. If we do not have sample metadata, then we cannot plate the sample. Right?

charles-cowart: I believe in this case, the issue is that if study_id is NULL, and sample_id is NULL, then it should be understood that this row is control data. After working with this code a bit, I think the best thing is for it to be rewritten at some point. The same method is being used to generate output for two different files (control and samples), and conditionally modifying orig_name values has been difficult without breaking other behavior. Also, the test data is stored as a string representation of two TSV files, one for control and another for samples; this is pretty hard to work with. There's other ways of testing for equality.

Currently, I have modified the Orig_name column in the samples data and removed the leading study id from each row ('1.'). I haven't modified the test control file data, but I have confirmed its Orig_name column doesn't have study IDs prepending it.

I believe what we can expect is that sample data will have an integer value prepending the other part of the string value, separated by a dot. The previous regex expression will remove those, while leaving the orig_ids that don't match alone. I think this is good enough for now.

labman/db/process.py
```diff
@@ -3514,7 +3526,8 @@ def generate_prep_information(self):
                         for k, v in df.orig_name.iteritems()]
             # If the orig_name is none (bc this item is a control),
             # use its content
```
AmandaBirmingham: Please update the comment ... as I understand it, this is no longer referring to "orig_name" but now to "orig_name2"

labman/db/process.py
```diff
@@ -3545,6 +3558,13 @@ def generate_prep_information(self):
                     'EXPERIMENT_DESIGN_DESCRIPTION'
             }
             df.rename(index=str, columns=mv, inplace=True)
+            # as orig_name2 has been transformed into Orig_name, and
+            # the original orig_name column has been used to generate
+            # sample name, there is no longer a purpose for the
+            # original orig_name column, hence drop it from the final
```
AmandaBirmingham: For (verbose) clarity, could you change "orig_name column has been used to generate sample name" to "orig_name column has been used to generate df.index, which will be used as the sample name"?

labman/db/process.py
```diff
@@ -3492,6 +3492,18 @@ def generate_prep_information(self):
                 result['center_project_name'] = ''
                 result['runid'] = ''
                 result['instrument_model'] = sequencing_run[0]
+                result['orig_name2'] = result['orig_name']
+
+                if result['orig_name2'] is not None and study_id is not None:
+                    # strip the prepended study id from orig_name2, but only
+                    # if this is a 'sample' row, and not a 'control' row.
+                    # (captured here w/orig_name2 and study_id not equaling
```
AmandaBirmingham: Could you change " 'sample' row " to " 'experimental sample' row "?

labman/db/process.py
```diff
@@ -3492,6 +3492,18 @@ def generate_prep_information(self):
...
+                    # None. This also prevents interference w/the population
+                    # of the DataFrame below, using the existing list
+                    # comprehension.
```
AmandaBirmingham: Could you change "Dataframe below" to "Dataframe index below"?

labman/db/process.py
charles-cowart: Will do!

labman/db/process.py
charles-cowart: sure thing!

labman/db/process.py
charles-cowart: done!

labman/db/process.py
charles-cowart: Gotcha.

charles-cowart:
An regular email containing the change to Orig_name column sent to Gail. @AmandaBirmingham and @wasade CCed. Contents post here as well:
[Issue429.xlsx](https://github.com/jdereus/labman/files/3087128/Issue429.xlsx)

charles-cowart:
@AmandaBirmingham I think that's everything requested. Is there anything else I can modify?

charles-cowart:
@AmandaBirmingham thanks!

PR #452, opened 2019-04-15
Remove Study ID from prep info file

charles-cowart:
This mod will remove the study_id from the beginning of each row in column
'Orig_ID', if there exist numbers at the beginning of the string that
match the proper Study ID. If not, it will leave the entry untouched.


charles-cowart:
Duplicate of https://github.com/jdereus/labman/pull/446

Issue #394, opened 2018-11-15
Change "labman" to "labcontrol" in developer-facing files


AmandaBirmingham:
See related issues  #175 Decide on new name to replace labman and #393 Change "Labman" to "LabControl" in user interface.

charles-cowart:
Made a pass replacing all instances of 'labman' and 'Labman' in the code with 'lab control' and 'Labcontrol', respectively. Changes have been pushed to 'Issue394' branch. Will flake8 before creating a PR.

For now, the reference to Jeff's GitHub project 'jdereus/labman' has been kept, until a new home for the project can be determined.

charles-cowart:
We'll need to make sure the labman references in https://github.com/jdereus/labman/blob/master/docs/index.rst are changed as well, since they're being used as the Help menu docs in LabControl.

charles-cowart:
Just to keep track, https://github.com/jdereus/labman/pull/459 also touched on this issue.

charles-cowart:
https://github.com/jdereus/labman/pull/534 Also touches on this issue.

AmandaBirmingham:
@charles-cowart  AFAIcan tell, the only place "labman" is left is in the name of our repo :)  Should we close this issue, or modify it to address the repo name itself?

charles-cowart:
Yeah, I've been thinking the same thing. I think only Jeff has ownership of
the repo; with that and other considerations it's a reasonable task just by
itself.

charles-cowart:
Closing this issue, as #534 removes all remaining instances of labman, save for the github project name itself. A new issue will be created for this migration.

PR #455, opened 2019-04-16
Partial name change to Labcontrol

charles-cowart:
'labman' source directory renamed to 'labcontrol'. All from and import references to 'labman.something' changed to 'labcontrol.something'. There are other references to labman that are yet to be changed to labcontrol; this should be just the minimum amount required to rename the labman source directory.


charles-cowart:
Just to reiterate, this doesn't include name changes in the database schema, etc. This was the minimum required to rename the source directory from 'labman' to 'labcontrol'.

AmandaBirmingham:
Relevant to #394

charles-cowart:
Thanks everyone! Who would like to merge?

AmandaBirmingham:
@charles-cowart  Go for it!

charles-cowart:
ty!

PR #464, opened 2019-04-18
Removing the copy manager allows autocomplete with paste to work

### LINKED ISSUE #354: Trigger autocomplete with paste as well as typing
Currently, pasting a sample id into the sample plating page does NOT trigger the autocomplete.  

To reproduce:
1) Go to the sample plating page and make a new plate
2) using the test database, choose the "Identification of the Microbes for Cannabis Soils" study
3) In one of the wells, *type in* (do not cut and paste) 640202
4) See that autocomplete correctly brings up the sample with this id:
screen shot 2018-10-02 at 10 29 48 am
5) In another well, *paste in* (do not type) 640202
6) See that autocomplete is not triggered, there is no option to select the correct sample with this id, and the sample is therefore shown as unknown:
screen shot 2018-10-02 at 10 30 09 am

wasade:
It seems that `CellExternalCopyManager` was prohibiting the autocomplete. `CellExternalCopyManager` is used to facilitate undo/redo. However, undo/redo appear to work without the copy manager so didn't see a reason to retain it.

Fixes #354.

autocomplete

### DIFF
--- a/labcontrol/gui/static/js/plateViewer.js
+++ b/labcontrol/gui/static/js/plateViewer.js
@@ -240,21 +240,12 @@ PlateViewer.prototype.initialize = function (rows, cols) {
     }
   });
 
-  var pluginOptions = {
-    clipboardCommandHandler: function(editCommand){
-      that._undoRedoBuffer.queueAndExecuteCommand.call(that._undoRedoBuffer,editCommand);
-    },
-    readOnlyMode : false,
-    includeHeaderWhenCopying : false
-  };
-
   this._frozenColumn = new Slick.Grid(this._frozenColumnTarget, this.frozenData,
                                       frozenColumn, frozenColumnOptions);
   this.grid = new Slick.Grid(this.target, this.data, sgCols, sgOptions);
 
   // don't select the active cell, otherwise cell navigation won't work
   this.grid.setSelectionModel(new Slick.CellSelectionModel({selectActiveCell: false}));
-  this.grid.registerPlugin(new Slick.CellExternalCopyManager(pluginOptions));
 
   // When a cell changes, update the server with the new cell information
   this.grid.onCellChange.subscribe(function(e, args) {


wasade:
The description of that parameter is "set to true and the plugin will take the name property from each column (which is usually what appears in your header) and put that as the first row of the text that's copied to the clipboard" ([src](https://github.com/mleibman/SlickGrid/pull/1109/commits/155c305b5908e1aea12931962047f881ceaa823c#diff-f8e858a1d2023f81760d35950dfb3381R25)). I don't believe that is needed? It's not clear to me what the benefit of retaining this plugin is since it looks like copy/paste works without it?

antgonza:
My _guess_ is that it was cause the source of the copy/paste could have the header; however, I'm not sure if that's an actual problem or request from wetlab.

wasade:
@charles-cowart, undo and redo appear to work without `CellExternalCopyManager`

antgonza:
@charles-cowart Now, that you say that, perhaps the CellExternalCopyManager needs some extra parameters to behave correctly ?

wasade:
So far there is no evidence that `CellExternalCopyManager` is needed. What does it do, that we cannot presently do, that warrants its inclusion?

charles-cowart:
@wasade @antgonza I'm for removing CellExternalCopyManager, if we can't demonstrate that it's changing behavior.

AmandaBirmingham:
I have zero knowledge about the workings (or not) of CellExternalCopyManager, but I dug through the commit history and verified that CellExternalCopyManager came in with PR #141 , "Add the ability to copy and paste".  That PR was written by @ElDeveloper, who thus might be able to offer an informed opinion on this ... @ElDeveloper , can you remember where the code in question came from, and/or any info about what it is there for?  Or what might blow up if we cut it out? :)

ElDeveloper:
Just tested this PR, without this object users can no longer copy/paste multiple cells simultaneously. Which is AFAIK desired and used functionality (namely when copying/pasting from external spreadsheets).

wasade:
Thanks, @ElDeveloper! What I think will be necessary then is to create a "mode" checkbox to enable/disable pasting from an external spreadsheet, where the enable/disable registers or unregisters the plugin.

charles-cowart:
@wasade Would you like to make this a requirement for this PR? Or should we make an entirely new issue out of this?

wasade:
I believe this should be a requirement of the PR, thanks for checking!

charles-cowart:
@wasade Not a huge deal, but maybe this title should be changed to 'WIP: Removing the copy manager allows autocomplete with paste to work'.

charles-cowart:
Closing, as this work has been superseded by incoming PRs from @fedarko.

PR #469, opened 2019-04-19
Adding CHANGELOG.md

charles-cowart:
Adding CHANGELOG.md, based on feedback from team.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,12 @@
+# LabControl changelog
+
+Version 0.2.0 (2019-MM-DD)
+--------------------------
+
+* Bug fixes for generated prep information files.
+* (ADD HERE)
+
+Version 0.1.0 (2018-11-16)
+--------------------------
+
+Initial soft-launch release.


charles-cowart:
@AmandaBirmingham Personally, I think everyone on the team should be able to edit the changelog when any change of significance has been made. It's probably a quicker way to get up to speed than going through the closed issues on GitHub.

antgonza:
@AmandaBirmingham, good question! IMOO the highlights of the issues being solved or big decisions that might make an impact in the future.

For example, a release CHANGELOG could become [a user email of new features in the release](https://forum.qiime2.org/t/qiime-2-2019-1-release-is-now-live/8123) or a [simple summary of the big changes](https://github.com/biocore/qiita/blob/master/CHANGELOG.md).

antgonza:
@charles-cowart, right but note that reviewers can also ask the PR author to add a line in the CHANGELOG; however, for Qiita I normally go over all the closed issued for that given release ...

charles-cowart:
@antgonza seems like it's a matter of whether you want an continuously updated list, or a summary of changes compiled and published at release. I can go either way; I've done both.

antgonza:
@charles-cowart, I _think_ we should start as a continuously updated list of good to knows, highlights of issues, etc until we have a formal release and then move to changes compiled and published at release but it's your call ...

charles-cowart:
@antgonza Agreed! I was thinking the same thing, actually.

PR #484, opened 2019-04-24
Modified setup to install slickgrid.plugins dir

charles-cowart:
Modified setup to install slickgrid.plugins directory


setup.py
```diff
@@ -47,6 +47,7 @@
                            'static/img/*',
                            'static/css/*',
                            'static/vendor/js/*',
+                           'static/vendor/js/slickgrid.plugins/*',
```
antgonza: Not sure if better but what about 'static/vendor/js/*/*'?

charles-cowart: Never tried that before but if it works, it sounds good! What happens if a new directory is doubly nested though? Is there a way to have it recursively grab ad infinitum?

ElDeveloper: In general I would advocate against a single wildcard as files not intended to be included could be shipped with the Python package. For example temporary files generated by text editors i.e. `slick.grid.core.js~`, etc. I would suggest that the wildcards be changed to specify what file types are needed from each folder. So change `static/img/*` to `static/img/*.png` and so on.

antgonza: Yes, after my comment I checked what are we doing in qiita and it looks like we add whatever libraries we need to the "root" folder, in this case `static/vendor/js/` and as @ElDeveloper said, we make sure to only add png/js/etc ...

charles-cowart: Thanks guys, will do!

charles-cowart: @antgonza I may be wrong, but I believe I had to keep the slickgrid.plugins specific entry, just because it seems that the wildcard isn't recursive, whether its alone or qualified with an extension.

antgonza:
@wasade approved and looks good to me ... merging!

charles-cowart:
Thanks!

PR #491, opened 2019-04-30
Beef up unit test to prep for fixing issue 457

AmandaBirmingham:
TL;DR: This PR doesn't fix any existing issues but rather prepares the unit tests to adequately test a future fix.  This is a very small change and I believe it could probably be reviewed by a single reviewer.

The existing unit test for NormalizationProcess._calculate_norm_vol did not exercise the min_vol clipping capability, which would make it impossible to adequately test a fix for issue 457.  I have added a test entry that exercises min_vol and comments explaining what each input entry tests.

Additionally, the unit test used 2-d (nested) arrays of concentrations as input.  I have examined how the real data is pulled from the database and transferred through the system, and it appears that instead it is treated as a 1-d array, as shown in this example:

Screen Shot 2019-04-29 at 4 19 06 PM

 Therefore I have changed the test input variable (exp_vols) to a 1-d array.


wasade:
Thanks @AmandaBirmingham! Will merge once CI finishes

AmandaBirmingham:
Since @wasade has approved merge, going ahead and merging it so I can move on to the actual issue fix :D

PR #492, opened 2019-04-30
Issue457

### LINKED ISSUE #457: Normalization workflow minor changes. (Volume units, default values)
### Minor changes to metagenomics input normalization workflow.

- The default parameter for the 'Min Volume' field in the input normalization workflow for metagenomics should be 25 and not 2.5 in order to match our Jupyter notebook default parameters. 
Screen Shot 2019-04-16 at 12 10 59 PM
Screen Shot 2019-04-16 at 12 15 33 PM

- All fields asking for volumes should display units in parentheses. (nL)

AmandaBirmingham:
Fixes #457 by changing default min_vol for gDNA normalization to 25 nL (instead of 2.5) and tweaking html text on normalization.html page to include units.

Pretty minimal changes, can probably get by with a single reviewer (y'all's call :)


labcontrol/db/process.py
```diff
@@ -1087,9 +1087,11 @@ def _calculate_norm_vol(dna_concs, ng=5, min_vol=2.5, max_vol=3500,
         ng : float, optional
             The amount of DNA to pool (ng). Default: 5
         min_vol : float, optional
-            The minimum volume to pool (nL). Default: 2.5
-        max_vol : float, optional
-            The maximum volume to pool (nL). Default: 3500
+            The minimum volume to pool (nL). Default: 25
+       max_vol : float, optional
```
wasade: I think the indent is missing a space here

wasade:
I believe a single reviewer on this is sufficient.

labcontrol/db/process.py
AmandaBirmingham: Good eye!  Thanks :)

Issue #431, opened 2019-04-03
Project_name column in prep sheet is empty in controls output


ackermag:
Should reflect the project that the blanks/controls refer to.

charles-cowart:
@ackermag I might need clarification on where this is occurring in the user interface.

charles-cowart:
To clarify, this is the 16S workflow's prep-information-file download (controls output file - 1 file of 2).

charles-cowart:
Confirmed Project_name column is empty in controls output. @ackermag says we can pull project_name 'Palmer_11479' from the sample_plate column ('Palmer_11479_Plate_1') for now, if need be (ignoring case where two projects are on the same plate.) Note samples file does have a populated Project name column ('Palmer GWAS, SG_B0014')

charles-cowart:
The code strongly suggests that study.study_alias is blank for controls:

study.study_alias AS project_name

result['project_name'] isn't touched in any way, until the column name is changed to 'Project_name'. I thought this was worth making a note of. At least for the short term, we'll probably have to copy and modify the value in the 'sample_plate' column.

AmandaBirmingham:
This issue and https://github.com/jdereus/labman/issues/483 are not the same, but they are *definitely* kissing cousins.  Please keep both in mind when attempting a fix.

AmandaBirmingham:
See #483 for question

AmandaBirmingham:
When solving this issue, we may want to determine if it intersects with/is affected by #387 "Add new `was_pooled` field to amplicon prep file for blanks" ... e.g., does it make sense to assign blanks/controls to a particular project if they weren't ever actually pooled (and thus sequenced)?  My understanding is that in the amplicon pipeline, usually only 2 blanks are pooled; see below.

https://github.com/jdereus/labman/blob/645c798e0094e9962973c2d111c4860b1e8e5b23/labcontrol/gui/handlers/process_handlers/pooling_process.py#L238-L243

charles-cowart:
@AmandaBirmingham Is this a question for @ackermag? My thought was that every control would be part of a project, but you're saying that if the control/blank wasn't plated, you'd prefer for it to remain blank in the controls prep sheet?

AmandaBirmingham:
[...]

PR #496, opened 2019-05-03
Issue445

### LINKED ISSUE #445: Consider merging preparation files
Mapping (Preparation) files are split into two separate files, one for blanks, and the other for samples. Wetlab prefers a single mapping file for both.

AmandaBirmingham:
Fixes #445 , putting all info from an (amplicon) run into the same prep file--at least for now.

Is necessary prerequisite to handling #431 .


labcontrol/db/tests/test_process.py
```diff
@@ -1830,16 +1830,14 @@ def test_generate_sample_sheet_unrecognized_assay_type(self):
...
-               controls_sheet_id: CONTROL_SAMPLES_PREP_EXAMPLE}
+        exp_key = 'Test Run.1'
+        exp = {exp_key: COMBINED_SAMPLES_PREP_EXAMPLE}
         self.assertEqual(len(obs), len(exp))
         # print observed output, so it can be captured and compared
         # with expected string.
-        print(obs[1])
-        self.assertEqual(obs[1], exp[1])
-        self.assertEqual(obs[controls_sheet_id], exp[controls_sheet_id])
+        print(obs[exp_key])
```
wasade: I think this is residual from testing :)

AmandaBirmingham: Ok; it was pre-existing (see previous line 1840) so I didn't mess with it, but I am happy to take it out.

PR #499, opened 2019-05-06
Fixes #456

### LINKED ISSUE #456: Lab Control Plate samples workflow character restrictions impede sample matching to sample identifier.
When plating iMSMS project (11326), which has column 'Tube_ID' as sample identifier, Lab Control GUI character restrictions make it impossible for one to match the 'Tube_ID' value because 'Tube_ID' is not character restricted. 

For more context, the iMSMS project provided plate maps with some samples with special characters. These were captured in the metadata column 'Tube_ID'.

Screen Shot 2019-04-16 at 11 10 47 AM

When trying to plate samples, the Lab Control GUI deletes the space in between  'Mag' and 'Bead' and fails to match the sample name in the 'Tube_ID' column of the metadata. 

Screen Shot 2019-04-16 at 11 11 05 AM

wasade:
I did not observe any oddities using the described suggested javascript modifications to the character restrictions.

Before:
issue_456

After:
issue_456_t2

### DIFF
--- a/labcontrol/gui/static/js/plateViewer.js
+++ b/labcontrol/gui/static/js/plateViewer.js
@@ -605,14 +605,12 @@ function SampleCellEditor(args) {
     if (state === null) {
       state = '';
     }
-    state = state.replace(/\s/g,'');
-    if (state.length === 0) {
+    
+    if (state.replace(/\s/g,'').length === 0) {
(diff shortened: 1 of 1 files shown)


wasade:
Am okay with a single review, thanks!

Issue #498, opened 2019-05-06
Prepare Sequencing Run allows users to mix amplicon and metagenomics pools, which leads to error downstream in sequencing run sample sheet preparation


AmandaBirmingham:
Currently, when selecting a sequencer that has more than one lane (and thus accepts more than one pool), nothing prevents the user from mixing amplicon pool(s) with shotgun (plate) pool(s).  Unfortunately, this mixing was not anticipated in the software development and causes problems at the sample sheet download step: if the shotgun pool is added first, the code assumes all pools are shotgun pools and throws an error when trying to trace back info to generate the sample sheet because the amplicon pools do not have normalized gdna records.  If the amplicon pool is added first, the code produces an amplicon-style sample sheet that is (probably?) not correct/useful for a shotgun pool.  See animated gif below demonstrating both use cases:

![issue498](https://user-images.githubusercontent.com/10677935/57261543-fe56b000-701c-11e9-8355-057c5ec119ec.gif)

AmandaBirmingham:
Note that the proximate reason for the incorrect behavior is 

https://github.com/jdereus/labman/blob/1b6163584d1a4a0e5222b200dfe6cda1253bf71f/labcontrol/db/process.py#L2713-L2715

Which only checks the composition type of the FIRST pool component when deciding whether the sequencing run is amplicon or shotgun.  However, the ultimate cause is that the code doesn't support having both kinds in the same pool (even if we DID detect it) so better to fix that.

Issue #497, opened 2019-05-06
Pool list page checkboxes allows user to add metagenomics (not amplicon) plate pools, which leads to error downstream in sequencing run sample sheet preparation


AmandaBirmingham:
See https://github.com/jdereus/labman/issues/420 for error produced

![issue497](https://user-images.githubusercontent.com/10677935/57271288-c6169800-7043-11e9-8254-f74bdd2d3859.gif)

Issue #444, opened 2019-04-09
Ensure content is downloadable.


charles-cowart:
"Latest mapping file downloaded is empty, however AGP mapping files in the past were downloaded with no trouble."

Review download buttons from existing plates, etc. and ensure content is downloadable. Where content is not downloadable, or is an invalid file, investigate as to the cause.

AmandaBirmingham:
We may have more info about what happened here, as a side-effect of today's meeting re miniPCR; below are my notes of what I heard from MacKenzie:

Ran through a single plate (HNRC) that was part of a bigger sequencing run, but only uploaded HNRC to LabControl, but didn't run Prepare Amplicon Sequencing Pool.  Note: were still offered options to download sample sheet and prep sheet, but they were corrupt.  

So: They were trying to download sample sheets and prep sheets *for a library plate pool*, not for an amplicon sequencing pool.  Betcha a dollar this is the ultimate cause of why the stuff that was downloaded was corrupt/empty.

AmandaBirmingham:
Hoo doggies--I tracked this one down.  The issue is reproducible: go to "Sequencing runs" > "Prepare sequencing run" and select *any amplicon library plate pool* (not an amplicon sequencing pool).  Create the sequencing run (which works just fine) and then go to "Sequencing runs" > "List" and click on the "Download Preparation Sheets" button next to the new sequencing run you just created.   A zip file will download but when you go to open it you will get an error like the one below:

<img width="490" alt="Screen Shot 2019-04-19 at 4 23 27 PM" src="https://user-images.githubusercontent.com/10677935/56447850-90418780-62bf-11e9-9137-fd740c64e588.png">

This is because the zip file is in fact empty, and that is because  

https://github.com/jdereus/labman/blob/645c798e0094e9962973c2d111c4860b1e8e5b23/labcontrol/db/process.py#L3244

... returns an empty dictionary, and THAT is because the giant sql statement in that function expects the pool composition associated with the sequencing run to be made up of OTHER pool compositions (as amplicon pools are, where you make library plate pools and then pool THEM together to get the pool for your sequencing run):

https://github.com/jdereus/labman/blob/645c798e0094e9962973c2d111c4860b1e8e5b23/labcontrol/db/process.py#L3374-L3380

And the reason the sql was written that was is that when we wrote it, we didn't know that the lab COULD use library plate pools for a sequencing run and just skip the amplicon sequencing pool step.  

[...]

Issue #420, opened 2019-04-02
Prepare Amplicon Sequencing Pool allows user to add metagenomics (not amplicon) plate pools, which leads to error downstream in sequencing run sample sheet preparation


charles-cowart:
Note: The original issue described in this comment, "Download Preparation Sheets button producing junk output" is a duplicate of https://github.com/jdereus/labman/issues/444 and has already been diagnosed there.  I have therefore changed the name of this issue to reflect the *sample sheet* related concerns raised in the comments.

-----------------------------------------------------------------------
Sequence Run List page has a 'download preparation sheets' button, which when pressed is initiating a download procedure for an invalid zip file 22 bytes long. Adjacent button for sample sheets is behaving correctly.

This appears specific to 'HNRC_HIV_4'. The other download preparation sheets buttons work as intended. 

Investigate the cause of this issue.

wasade:
Unsure if related, but it is possible to add plates for sequencing that don't appear to have been normalized. It seems that the plates listed are not assured to be valid. Note that, for recreation, I'm adding plates that were created following the name duplication in #479 -- that may or may not be related. It seems that downloading a sample sheet is possible when using one of the test pools. 

![sample_sheet_exception](https://user-images.githubusercontent.com/474290/56619418-51227780-65da-11e9-9031-7194b336111f.gif)


```bash
AN ERROR HAS OCCURED!
Please copy the following into an issue in our github repository

Error
'PoolComposition' object has no attribute 'normalized_gdna_composition'

Traceback
Traceback (most recent call last): 
File "/Users/dtmcdonald/miniconda3/envs/labcontrol/lib/python3.7/site-packages/tornado/web.py", line 1697, in _execute result = method(*self.path_args, **self.path_kwargs) 
File "/Users/dtmcdonald/miniconda3/envs/labcontrol/lib/python3.7/site-packages/tornado/web.py", line 3174, in wrapper return method(self, *args, **kwargs) 
File "/Users/dtmcdonald/ResearchWork/software/labman/labcontrol/gui/handlers/process_handlers/sequencing_process.py", line 59, in get text = process.generate_sample_sheet() 
File "/Users/dtmcdonald/ResearchWork/software/labman/labcontrol/db/process.py", line 3240, in generate_sample_sheet return self._generate_shotgun_sample_sheet() 
[...]

PR #500, opened 2019-05-07
Issue420

AmandaBirmingham:
While investigating the causes of #420 "Prepare Amplicon Sequencing Pool allows user to add metagenomics (not amplicon) plate pools, which leads to error downstream in sequencing run sample sheet preparation", I also found issues #497 "Pool list page checkboxes allows user to add metagenomics (not amplicon) plate pools, which leads to error downstream in sequencing run sample sheet preparation" (very similar to 420) and #498 "Prepare Sequencing Run allows users to mix amplicon and metagenomics pools, which leads to error downstream in sequencing run sample sheet preparation".  Basically all of these are related to the fact that although pools can be made up of any arbitrary contents (including any combination of amplicon library preps, shotgun library preps, and/or other pools), **it isn't currently sensical to *sequence* such a pool** as we don't know what the downstream outputs (sample sheets and/or prep sheets) for arbitrary pools (made up of mixtures of types of inputs) should look like.

For now, I solved all three of these issues by limiting which kinds of inputs are available to the user when creating an amplicon sequencing pool (can only input amplicon plate pools), a shotgun sequencing run (can only input shotgun plate pools), or an amplicon sequencing run (can only input amplicon sequencing pools).  In the future, we may decide that (a) there are realistic use-cases for mixing the types of inputs that go into a pool AND (b) we understand what the downstream outputs for sequencing runs of arbitrary pools should look like.  If and when that day comes, we will want to take out the limitations I have added here.

Note that these limitations *prevent* the case described in issue #444 "Modify generate_prep_information to correctly create prep sheets for sequencing runs based on library plate pools" from occurring, thus making it a low priority to fix.

Also note that there are some known code smells here ... for example, an examination of what kinds of library prep components go into a pool is now happening in `util.py`'s `get_pool_listing` as well as in `process.py`'s `SequencingProcess.create()` (the latter of which I am starting to think is the wrong place for it and is related to the bug for issue #498), and the definition of "amplicon_sequencing" pools as given in `PoolListHandler.get()` would also catch additional kinds of pools if we add additional protocols.  Fixing these smells will take a bit of thoughtful refactoring of how and where we identify different assay types; if this PR gets approved, I'll add an issue for that.

### DIFF
--- a/labcontrol/gui/test/test_pool.py
+++ b/labcontrol/gui/test/test_pool.py
@@ -13,20 +13,65 @@ from labcontrol.gui.testing import TestHandlerBase
 
 
 class TestPoolHandlers(TestHandlerBase):
-    def test_get_plate_list_handler(self):
-        response = self.get('/pool_list')
+    def test_get_pool_list_handler_all(self):
+        response = self.get('/pool_list/all/')
         self.assertEqual(response.code, 200)
         obs = json_decode(response.body)
         self.assertCountEqual(obs.keys(), ['data'])
         obs_data = obs['data']
         self.assertEqual(len(obs_data), 6)
-        self.assertEqual(obs_data[0], [1, 'Test Pool from Plate 1', True, 1])
-        self.assertEqual(obs_data[1], [2, 'Test sequencing pool 1', False, 2])
+        self.assertEqual(obs_data[0], [1, 'Test Pool from Plate 1',
+                                       True, True, 1])
+        self.assertEqual(obs_data[1], [2, 'Test sequencing pool 1',
+                                       False, False, 2])
         self.assertEqual(obs_data[2], [3, 'Test pool from Shotgun plates 1-4',
-                                       True, 3])
-        self.assertEqual(obs_data[3], [4, 'Test Pool from Plate 2', True, 4])
-        self.assertEqual(obs_data[4], [5, 'Test Pool from Plate 3', True, 5])
-        self.assertEqual(obs_data[5], [6, 'Test Pool from Plate 4', True, 6])
+                                       True, False, 3])
+        self.assertEqual(obs_data[3], [4, 'Test Pool from Plate 2',
+                                       True, True, 4])
+        self.assertEqual(obs_data[4], [5, 'Test Pool from Plate 3',
+                                       True, True, 5])
+        self.assertEqual(obs_data[5], [6, 'Test Pool from Plate 4',
+                                       True, True, 6])
+
+    def test_get_pool_list_handler_amplicon_plate(self):
+        response = self.get('/pool_list/amplicon_plate/')
+        self.assertEqual(response.code, 200)
+        obs = json_decode(response.body)
+        self.assertCountEqual(obs.keys(), ['data'])
+        obs_data = obs['data']
+        self.assertEqual(len(obs_data), 4)
+        self.assertEqual(obs_data[0], [1, 'Test Pool from Plate 1',
+                                       True, True, 1])
+        self.assertEqual(obs_data[1], [4, 'Test Pool from Plate 2',
+                                       True, True, 4])
+        self.assertEqual(obs_data[2], [5, 'Test Pool from Plate 3',
+                                       True, True, 5])
+        self.assertEqual(obs_data[3], [6, 'Test Pool from Plate 4',
+                                       True, True, 6])
+
+    def test_get_pool_list_handler_amplicon_sequencing(self):
+        response = self.get('/pool_list/amplicon_sequencing/')
+        self.assertEqual(response.code, 200)
+        obs = json_decode(response.body)
+        self.assertCountEqual(obs.keys(), ['data'])
+        obs_data = obs['data']
+        self.assertEqual(len(obs_data), 1)
+        self.assertEqual(obs_data[0], [2, 'Test sequencing pool 1',
+                                       False, False, 2])
+
+    def test_get_pool_list_handler_metagenomics_plate(self):
+        response = self.get('/pool_list/metagenomics_plate/')
+        self.assertEqual(response.code, 200)
+        obs = json_decode(response.body)
+        self.assertCountEqual(obs.keys(), ['data'])
+        obs_data = obs['data']
+        self.assertEqual(len(obs_data), 1)
+        self.assertEqual(obs_data[0], [3, 'Test pool from Shotgun plates 1-4',
+                                       True, False, 3])
+
+    def test_get_pool_list_handler_unknown(self):
+        response = self.get('/pool_list/minipcr/')
+        self.assertEqual(response.code, 500)
 
 
 if __name__ == '__main__':
--- a/labcontrol/gui/test/test_pool.py
+++ b/labcontrol/gui/test/test_pool.py
@@ -59,8 +59,8 @@ class TestPoolHandlers(TestHandlerBase):
         self.assertEqual(obs_data[0], [2, 'Test sequencing pool 1',
                                        False, False, 2])
 
-    def test_get_pool_list_handler_metagenomics_plate(self):
-        response = self.get('/pool_list/metagenomics_plate/')
+    def test_get_pool_list_handler_shotgun_plate(self):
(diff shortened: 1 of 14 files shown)


AmandaBirmingham:
Because this PR includes various front-end fixes that are difficult to grok from the code alone, I'm adding comments below about what each of those fixes look like.

AmandaBirmingham:
#420 "Prepare Amplicon Sequencing Pool allows user to add metagenomics (not amplicon) plate pools, which leads to error downstream in sequencing run sample sheet preparation"

Before PR:

Screen Shot 2019-05-06 at 9 33 56 PM

Screen Shot 2019-05-06 at 9 34 02 PM


After PR:

Screen Shot 2019-05-06 at 3 27 08 PM

Note that header clarifies this page is only for making an *amplicon* sequencing pool

Screen Shot 2019-05-06 at 3 27 16 PM

Note that only amplicon plate pools are available as inputs

AmandaBirmingham:
#497 "Pool list page checkboxes allows user to add metagenomics (not amplicon) plate pools, which leads to error downstream in sequencing run sample sheet preparation" 

Before PR:

Screen Shot 2019-05-06 at 9 41 16 PM

Screen Shot 2019-05-06 at 9 41 29 PM

After PR:

Screen Shot 2019-05-06 at 3 25 39 PM

Note that pools that are NOT amplicon plate pools have grayed-out checkboxes because they cannot "legally" be used to make amplicon sequencing pools.

Screen Shot 2019-05-06 at 3 25 54 PM

Note that button has been relabeled to "Prepare amplicon sequencing pool" to match change to header on the corresponding page.

AmandaBirmingham:
#498 "Prepare Sequencing Run allows users to mix amplicon and metagenomics pools, which leads to error downstream in sequencing run sample sheet preparation".

Before PR:

Screen Shot 2019-05-06 at 9 45 18 PM

Screen Shot 2019-05-06 at 9 45 57 PM

Screen Shot 2019-05-06 at 9 46 12 PM

After PR: 

Screen Shot 2019-05-06 at 10 03 51 PM

Now there are separate menu items for making a run from amplicon pools vs making a run from shotgun pools.  I went with this because I think it will be clear to the user, it can be supported with essentially no code duplication (just different input variables into the same html template page), and because having one single page *switch back and forth* between shotgun and amplicon input options and clear out state between them (a la `library_pooling.html`) has proven to be an invitation to bugs.

Screen Shot 2019-05-06 at 3 27 50 PM

Screen Shot 2019-05-06 at 3 28 02 PM

Note that if you pick the "prepare amplicon sequencing run" menu option, the header indicates this page is for amplicon inputs only and the pools you can choose from are ONLY the amplicon sequencing pools (so issue #444 can no longer be caused). Conversely, note that if you pick the "prepare shotgun sequencing run" menu option, the header indicates this page is for shotgun inputs only and the pools you can choose from are ONLY the shotgun plate pools

Screen Shot 2019-05-06 at 10 04 23 PM
 
Screen Shot 2019-05-06 at 3 28 24 PM

charles-cowart:
> Note that pools that are NOT amplicon plate pools have grayed-out checkboxes because they cannot "legally" be used to make amplicon sequencing pools.

Would it be more clear to remove the non-amplicon plate pools?

charles-cowart:
> and because having one single page switch back and forth between shotgun and amplicon input options and clear out state between them (a la library_pooling.html) has proven to be an invitation to bugs.
+1

AmandaBirmingham:
> > Note that pools that are NOT amplicon plate pools have grayed-out checkboxes because they cannot "legally" be used to make amplicon sequencing pools.
> 
> Would it be more clear to remove the non-amplicon plate pools?

No, because this is occurring on the pool list page, which exists primarily for the purpose of allowing the user can see all the pools (of all kinds) that have been made.  Allowing the user to click on checkboxes next to them and shortcut amplicon sequencing pool creation is an add-on rather than the main goal (an add-on that only applies to certain kinds of pools :( ). 
 
There are a couple of other options besides graying out the checkboxes … we could make the user pick a pool type (as we make them pick the plate type in the plate list page), but we don’t currently have explicit pool types the way we have explicit plate types and that whole “switching types on the same page and clearing state” thing hasn’t worked out well for us in the past.  Alternately, we could entirely remove the checkboxes for the pools that can’t be included in amplicon sequencing pools, but I thought that would be MORE confusing to the user than graying them out (perhaps others disagree …).

labcontrol/gui/test/test_pool.py
```diff
@@ -13,20 +13,65 @@
 
 
 class TestPoolHandlers(TestHandlerBase):
-    def test_get_plate_list_handler(self):
-        response = self.get('/pool_list')
+    def test_get_pool_list_handler_all(self):
+        response = self.get('/pool_list/all/')
```
charles-cowart: The '/all/' component of the resource string seems redundant - '/pool_list/' by itself, and with no further qualifying resource should display all resources by default, IMHO.

charles-cowart:
Yeah, picking a pool type doesn't seem like a good route. What about graying out the text too? That might be less confusing than just graying out the box, but that's just my thinking.

labcontrol/gui/test/test_pool.py
wasade: On the other hand, the internal behavior of the handler would be implicit. I'd be more in favor of explicit behavior

wasade:
This is a thorough PR, and the additional contextual comments in the code are very beneficial. I'm strongly in favor of the explicitness of the approach, and the conservative behavior on handling pools. And I agree that the code smells can be deferred to a subsequent issue

AmandaBirmingham:
Thanks, @wasade ! Do you have any opinions on the discussion above about the checkboxes? (see https://github.com/jdereus/labman/pull/500#issuecomment-489915946 and https://github.com/jdereus/labman/pull/500#issuecomment-489917864 ).  

I would prefer not to gray out the whole row for non-amplicon-plate pools, as I worry that would make the user think there was something wrong with them (and the user might well be on this page to do something completely unrelated to making amplicon sequencing pools from amplicon plate pools).  Also, more pragmatically, it would require me to make more javascript changes :D   

However, I do acknowledge that the grayed out checkboxes are not a great UI approach, so if anyone has a (preferably low-effort :) third option, I'm open to it ...

labcontrol/gui/test/test_pool.py
AmandaBirmingham: @charles-cowart Given that @wasade likes the explicit behavior--and given that I actually don't know how to tweak the tornado handler regex to do both the explicit behavior and the implicit behavior in the same handler set-up :( -- would you be willing to table this idea?

wasade:
> > > Note that pools that are NOT amplicon plate pools have grayed-out checkboxes because they cannot "legally" be used to make amplicon sequencing pools.
> > 
> > 
> > Would it be more clear to remove the non-amplicon plate pools?
> 
> No, because this is occurring on the pool list page, which exists primarily for the purpose of allowing the user can see all the pools (of all kinds) that have been made. Allowing the user to click on checkboxes next to them and shortcut amplicon sequencing pool creation is an add-on rather than the main goal (an add-on that only applies to certain kinds of pools :( ).

I agree here, while also agreeing that greying out the checkbox is not the most ideal as it can sometimes be difficult to differentiate.

Would it be complex to replace the greyed out checkbox with "N/A" ? One other option would be to add an italicized message at the bottom of the modal stating "_Greyed out check boxes indicate a pool that is not applicable (e.g., amplicon with shotgun)_". (or if "N/A" is preferable, making the italicized statement to that effect).

AmandaBirmingham:
Based on https://github.com/jdereus/labman/pull/500#issuecomment-490249289, here's what the pool list page looks like now.  Please indicate if this works for y'all ... :)

Screen Shot 2019-05-07 at 1 58 43 PM

AmandaBirmingham:
Note that I've also merged in the small change to master that was made after this PR was created.

labcontrol/gui/test/test_pool.py
charles-cowart: definitely - not a problem.

Issue #483, opened 2019-04-24
Metagenomics Sample Sheet generation assigns project blank wells to a separate folder. 


RodolfoSalido:
Sample Sheet assigns a value of 'Controls'  in 'Sample_project' column to Blank wells. This leads to reads from Blank wells from distinct projects to get demultiplexed into one common 'Controls' folder per sequencing run independent of their respective project folders.

<img width="1354" alt="Screen Shot 2019-04-24 at 2 12 43 PM" src="https://user-images.githubusercontent.com/13193891/56694682-1b45c780-669c-11e9-9f7c-1e54d1b37cc4.png">

AmandaBirmingham:
@RodolfoSalido  Thank you very much for pointing this out! What value do you want to go into this column of the sample sheet for blanks?

AmandaBirmingham:
It is coming back to me a little: this Sample_project column was the source of *huge* amounts of discussion and requirements gathering (see https://github.com/jdereus/labman/issues/204) .  Specifically, https://github.com/jdereus/labman/issues/204#issuecomment-387128070 from @tanaes said 

> OK, after chatting about this with some folks, it seems like the best option vis-a-vis the study sheet is to have any of the controls on a sequencing run end up in a 'Controls' demultiplex folder after BCL2Fastq. We don't necessarily need to make this an actual Qiita study, but that would enable a uniform place to access control samples downstream.

Has this conclusion changed?

RodolfoSalido:
I had a hunch that there had to be a discussion about this because it appeared to be designed.  

I raised the issue because Greg thought it could be  problematic to have all blanks pooled into one folder per sequencing run. I’ve cc’ed him so he can elaborate. 

-Rodolfo

RodolfoSalido:
Controls cannot be an output folder for metagenomics... the blanks are associated with a specific extraction and project. And these controls are project based so sometimes they are fecal and sometimes they are skin. We know there is well to well contamination so controls for each project are different.

AmandaBirmingham:
This issue and https://github.com/jdereus/labman/issues/431 are not the same, but they are *definitely* kissing cousins.  Please keep both in mind when attempting a fix.

AmandaBirmingham:
@RodolfoSalido  wrote:

> Controls cannot be an output folder for metagenomics...

Ok, so this means the specifications for this code have changed :(, and the stakeholders on this need to **re-have** the conversation about assigning sample project name for blanks and controls. The above comment adds:

[...]

PR #501, opened 2019-05-07
Fixes #431

### LINKED ISSUE #431: Project_name column in prep sheet is empty in controls output

AmandaBirmingham:
Also provides supporting code that can also be used in future for #483.

@charles-cowart note that the additions made here to `_generate_amplicon_prep_information` should also make it into the new `_generate_metagenomics_prep_information`.


labcontrol/db/process.py
```diff
@@ -2998,6 +2998,95 @@ def _format_sample_sheet_comments(principal_investigator=None,
...
+            # get unique project names for the part of df defined in the mask
+            curr_unique_projnames = (input_df[plate_non_controls_mask]
+                                  [projname_col_name].unique())
+            if len(curr_unique_projnames) != 1:
+                # Note that we don't error out the first time we find a
+                # plate that doesn't meet expectations; instead we continue to
+                # run through all the plates and identify ALL those that don't
+                # meet expectations.  This way the user can correct all of them
+                # at once.
+                curr_err_msg = "Expected one unique value for plate '{0}' " \
```
wasade: There is a heinous edge case here: it is possible that `len(curr_unique_projnames) == 0`. I don't believe it would break the formation of the error message, but the nature of the error is different.

labcontrol/db/process.py
```diff
@@ -2998,6 +2998,95 @@ def _format_sample_sheet_comments(principal_investigator=None,
...
+                problem_plate_messages.append(curr_err_msg)
+            else:
+                # create a mask to define all the rows for this plate where the
+                # projname IS the control value (None); ~ "nots" a whole series
+                plate_controls_mask = plate_mask & (~non_controls_mask)
+
+                # ok to just take first non-control projname because we
+                # verified above there is only one value there anyway
+                input_df.loc[plate_controls_mask, projname_col_name] = \
+                    curr_unique_projnames[0]
```
wasade: ....however, if `len(curr_unique_projnames) == 0`, this would throw an `IndexError`. 

If I'm thinking about this correctly, I believe this edge case could result if a plate entirely of controls is run.

labcontrol/db/process.py
```diff
@@ -2998,6 +2998,95 @@ def _format_sample_sheet_comments(principal_investigator=None,
...
+        ------
+        ValueError
+            If any plate contains experimental samples from more (or fewer)
+            than one project.
+        """
+
+        problem_plate_messages = []
+
+        # create a mask to define all the NON-control rows for this plate
+        non_controls_mask = input_df[projname_col_name].notnull()
```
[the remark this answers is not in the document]
AmandaBirmingham: Yes, they would implicitly be treated as controls--and would be assigned to the same project as everything else on their plate, which is exactly what the lab wanted (e.g. in the case of the AGP samples that were not in qiita but were plated on an AGP plate and that were *previously* given a project name of "" instead of the AGP project name of the rest of the samples on the plate).

labcontrol/db/process.py
```diff
@@ -2998,6 +2998,95 @@ def _format_sample_sheet_comments(principal_investigator=None,
...
+        for curr_unique_plate in unique_plates:
+            # create a mask to define all the rows for this plate
+            plate_mask = input_df[plate_col_name] == curr_unique_plate
+
+            # create a mask to define all the rows for this plate where the
+            # project name is NOT the control value (None)
+            plate_non_controls_mask = plate_mask & non_controls_mask
+
+            # get unique project names for the part of df defined in the mask
+            curr_unique_projnames = (input_df[plate_non_controls_mask]
```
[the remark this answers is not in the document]
AmandaBirmingham: Will do

labcontrol/db/process.py
AmandaBirmingham: Nope, it won't throw an IndexError in this case--that situation is already taken care of above with the check `len(curr_unique_projnames) != 1` ... if the length is zero, you end up in the error handling and not ever in the `else`.  I have a test to verify this; will check it in in a moment :)

labcontrol/db/process.py
AmandaBirmingham: That edge case has already been anticipated and is handled by this logic. The error string recorded in that case looks like, e.g., "Expected one unique value for plate 'Test plate 3' but received 0: ".  I agree that the extra colon at the end is a bit ugly but I don't think it is worth refactoring for an edge case that we keep being told will never occur ... :)

labcontrol/db/process.py
wasade: Oh, wow, sorry. Yes, I misread the flow of logic there.

labcontrol/db/process.py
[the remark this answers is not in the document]
wasade: Thanks

AmandaBirmingham:
Ping @charles-cowart ... do you have an estimate on when you will be available to review this PR?  To sweeten the pot :), I have a fix for #483 ready to go that depends on this PR ... :D

charles-cowart:
@AmandaBirmingham Sorry for the delay, you're ready to go! :)

Issue #324, opened 2018-08-16
Extend Preparation Info file to have an Ext_robot_EP column and an Ext_robot_KF column instead of a single Extraction_robot column


AmandaBirmingham:
Requested by Greg; Gail agrees.

From: Gail Ackermann 
Date: Friday, August 10, 2018 at 5:22 PM

Absolutely fine with me...

On Fri, Aug 10, 2018 at 5:02 PM, Birmingham, Amanda wrote:
Hi, Gail,

Greg has requested a small change to the amplicon prep info file (to take effect once Labman goes live): instead of a single “Extraction_robot” column, there will be (in its place) two columns called “Ext_robot_EP” and “Ext_robot_KF”, holding the epmotion and kingfisher robots used in extraction, respectively.  Since you consume the prep info file, please let me know whether you are ok with this change; I don’t want to make modifications if they would disrupt your downstream workflow.

PR #502, opened 2019-05-08
WIP Gen shotgun prep files

charles-cowart:
Initial query and code for shotgun prep info file generation. My assumption is that we will be tweaking the columns, format of the output, etc. and getting approval from Gail, etc. before merging.


charles-cowart:
Partial implementation of final output using DataFrames instead of CSV for munging. Will pick it up in the morning.

wasade:
@charles-cowart, note I think this needs to be rebased

labcontrol/db/process.py
```diff
@@ -3623,7 +3629,8 @@ def _generate_amplicon_prep_information(self):
             for result in TRN.execute_fetchindex():
                 result = dict(result)
                 study_id = result.pop('study_id')
-                sid = result.pop('sample_id')
+                # sid unused when generating merged sample/controls prep file
+                # sid = result.pop('sample_id')
```
wasade: Can this be removed?

labcontrol/db/process.py
```diff
@@ -3783,3 +3790,407 @@ def _generate_amplicon_prep_information(self):
...
+        A support method for Metagenomics prep info file generation. This
+        method is only called by _generate_metagenomics_prep_information().
+        Gathers metadata used by above method and performs initial munging
+        for clarity.
+
+        Returns
+        -------
+        list: dict, each one representing a row of results.
+
+        #note: this fetchall() seemed appropriate, as we only expect to return
```
wasade: Should this be a notes section? eg:

```python
def foo(bar):
    """One liner

    Detailed description

    Parameters
    ----------
    stuff : stuff
        foo

    Returns
    -------
    list : dict ...

    Notes
    -----
    This fetchall() seemed appropriate...
    """

...our should those notes be moved to a comments block outside of the docstring?

**Comment by charles-cowart** [comment]
Sample output from the generate_metagenomics_prep_info_file() method. This is currently being used as the expected result for testing.

[shotgun_prep_info_sample_output.txt](https://github.com/jdereus/labman/files/3164296/shotgun_prep_info_sample_output.txt)

File: labcontrol/db/process.py [hunk 3]
```diff
@@ -3783,3 +3790,407 @@ def _generate_amplicon_prep_information(self):
...
+        #several hundred results at most. This allows us to capture the results
+        #and clean them up before handing them off. This also allows us to
+        #refactor this query in time without touching the rest of the code.
+        """
+        inst_mdl, equipment, reagent = self._get_additional_prep_metadata()
+
+        sql = """
```
wasade: Non-blocking, this makes me cringe, but I recognize that it may be a necessary evil right now. That, or I'm just not a strong database developer in which case, please ignore my novice cringing.

labcontrol/db/process.py
```diff
@@ -3783,3 +3790,407 @@ def _generate_amplicon_prep_information(self):
...
+                d['primer_date_i7'] =\
+                    d['primer_date_i7'].strftime(Process.get_date_format())
+
+                # instrument_model remains the same across all rows in this
+                # query.
+                d['instrument_model'] = inst_mdl
+
+                # note that the correct term is 'Kapa', not 'kappa'.
+                id = d['kappa_hyper_plus_kit_id']
+                d['kapa_hyper_plus_kit_lot'] = reagent[id]['external_lot_id']
```
wasade: Both `kappa` (line 3966) and `kapa` (line 3967) are used. Would it be horrendous to standardize on one right now (even if incorrect for the time being)?

labcontrol/db/process.py
```diff
@@ -3819,3 +3826,407 @@ def _generate_amplicon_prep_information(self):
...
+
+                # Note extraction_kit_id references (as in foreign-key)
+                # reagent_composition(reagent_composition_id).
+                id = d['extraction_kit_id']
+                d['extraction_kit_lot'] = reagent[id]['external_lot_id']
+
+                id = d['epmotion_tool_id']
+                d['epmotion_tool_name'] = equipment[id]['external_id']
+
+                # for now, platform is hard-coded to 'Illumina'
```
wasade: Can #507 be noted here in the comments?

labcontrol/db/process.py
```diff
@@ -3819,3 +3826,407 @@ def _generate_amplicon_prep_information(self):
...
+                                             d['orig_name2'])
+
+            return results
+
+    def _generate_metagenomics_prep_information(self):
+        """Generates prep information for Metagenomics workflows
+
+        An internal method used to implement the generation of prep information
+        files for Metagenomics workflows. This method is called by
+        generate_prep_information() only. This method is not currently
```
wasade: It seems like there is an implementation below :)

labcontrol/db/process.py
```diff
@@ -3819,3 +3826,407 @@ def _generate_amplicon_prep_information(self):
...
+
+        data = {}
+
+        for item in results:
+            # format well
+            well = []
+            col = item['col_num']
+            row = item['row_num']
+            while row:
+                row, rem = divmod(row - 1, 26)
```
wasade: Weird things I believe would happen if `row` were `<= 0`. The database does not appear to enforce that the row value is `[1, inf)`:

https://github.com/jdereus/labman/blob/313d006ccad1d9111d218b7ce645eb6d10fe37a9/labcontrol/db/support_files/db_patch.sql#L242-L249

Would it be possible to place an assertion on the valid range for both `row` and `col` out of an abundance of caution?

labcontrol/db/process.py
```diff
@@ -3819,3 +3826,407 @@ def _generate_amplicon_prep_information(self):
...
+            curr_prep_sheet_id = self.run_name
+            if curr_prep_sheet_id not in data:
+                data[curr_prep_sheet_id] = {}
+
+            content = item['content']
+
+            # adding item to the data (organized by prep_sheet_id and
+            # content string.
+            if content in data[curr_prep_sheet_id]:
+                raise ValueError("'%s' appears more than once" % content)
```
wasade: ~~I believe `content` is a large bolus of text here, which may make digestion of the traceback difficult. Would it be more helpful for an error to dump `curr_prep_sheet_id`? (I'm really not sure)~~

nm, it looks like this is a sample composition? Would it make sense to provide in the traceback both the `curr_prep_sheet_id` and `content`?

labcontrol/db/process.py
```diff
@@ -3819,3 +3826,407 @@ def _generate_amplicon_prep_information(self):
...
+            # Set the project column value for each non-experimental sample to
+            # the value of the project name for the (single) qiita study on
+            # that sample's plate.
+            prep_sheet =\
+                self._set_control_values_to_plate_value(prep_sheet,
+                                                        'sample_plate',
+                                                        'project_name')
+
+            # mapping keys to expected names for columns in the final output
+            mv = {""
```
wasade: Is the empty string needed here?

labcontrol/db/process.py
```diff
@@ -3819,3 +3826,407 @@ def _generate_amplicon_prep_information(self):
...
+                  "barcode_i7": "index",
+                  "barcode_i5": "index2",
+                  "primer_plate_i7": "i7_Primer_Plate",
+                  "primer_plate_i5": "i5_Primer_Plate",
+                  "primer_date_i7": "i7_Primer_date",
+                  "primer_date_i5": "i5_Primer_date",
+                  "experiment_design_description":
+                      "EXPERIMENT_DESIGN_DESCRIPTION",
+                  "instrument_model": "INSTRUMENT_MODEL",
+                  "kapa_hyper_plus_kit_lot": "KapaHyperPlusKit_lot",
```
wasade: Would it be possible to add an inline comment noting #503? It would be wonderful for #503 to have a permalink to this line as well, may reduce some confusion later when resolving that item.

labcontrol/db/process.py
```diff
@@ -3819,3 +3826,407 @@ def _generate_amplicon_prep_information(self):
...
+                  "stub_lot_id": "Stub_lot",
+                  "platform": "PLATFORM",
+                  "sequencing_method": "sequencing_meth",
```
wasade: A google image search for "sequencing meth" is not as exciting as I had hoped. Slack's `/giphy` on the other hand...

labcontrol/db/process.py
```diff
@@ -3819,3 +3826,407 @@ def _generate_amplicon_prep_information(self):
...
+                  "extraction_kit_lot": "ExtractionKit_lot",
+                  "epmotion_tool_name": "TM1000_8_tool",
+                  "i5_index_id": "i5_Index_ID",
+                  "i7_index_id": "i7_Index_ID",
+                  "library_construction_protocol":
+                      "LIBRARY_CONSTRUCTION_PROTOCOL"}
+            prep_sheet = prep_sheet.rename(columns=mv)
+
+            # Synthesize new columns
+            # prep_sheet['Orig_Sample_ID'] = prep_sheet['Orig_name']
```
wasade: Can this be removed please?

labcontrol/db/tests/test_process.py
```diff
@@ -1935,20 +1937,19 @@ def test_generate_amplicon_prep_information(self):
...
-        exp_err = "Prep file generation is not implemented for " \
-                  "Metagenomics assays."
         tester = SequencingProcess(2)
-        with self.assertRaisesRegex(ValueError, exp_err):
-            tester.generate_prep_information()
+        obs = tester.generate_prep_information()
+        exp_key = 'TestShotgunRun1'
+        exp = {exp_key: COMBINED_SAMPLES_METAGENOMICS_PREP_EXAMPLE}
+        self.assertEqual(len(obs), len(exp))
+        self.assertEqual(obs[exp_key], exp[exp_key])
```
wasade: Would it make sense for this to be `self.assertEqual(obs, exp)`?

labcontrol/db/process.py
charles-cowart: Nice catch! I wonder if that came back in with the auto-merge; one moment...

labcontrol/db/process.py
wasade: ...or should constraints be added elsewhere?

labcontrol/db/process.py
charles-cowart: Definitely! Will do.

labcontrol/db/process.py
AmandaBirmingham: Might want to consider logging this as a future issue rather than changing it here, because this very same logic exists in `_generate_amplicon_prep_information` and should probably be centralized before we start fixing it :)

labcontrol/db/process.py
charles-cowart: Unfortunately, it's referenced as 'kappa' in its database column, so if we truly wanted to go one way right now, we'd probably be better off staying as 'kappa'. That would also make the column name in the prep file 'kappa' also, unless we want to break standardization for the very final product. That would be my vote.

labcontrol/db/process.py
charles-cowart: Will do.

labcontrol/db/process.py
```diff
@@ -3819,3 +3826,407 @@ def _generate_amplicon_prep_information(self):
...
+                    samplecp.content,
+                    sampleplate.external_id AS sample_plate,
+                    platingprpr.run_personnel_id AS plating,
+                    gdnaextractpr.extraction_kit_id,
+                    gdnaextractpr.epmotion_robot_id AS gepmotion_robot_id,
+                    gdnaextractpr.epmotion_tool_id,
+                    gdnaextractpr.kingfisher_robot_id,
+                    libpreppr.kappa_hyper_plus_kit_id,
+                    libpreppr.stub_lot_id,
+                    libpreppr.normalization_process_id,
```
AmandaBirmingham: Is this id being used anywhere?

labcontrol/db/process.py
charles-cowart: 👍

labcontrol/db/process.py
```diff
@@ -3819,3 +3826,407 @@ def _generate_amplicon_prep_information(self):
...
+                FROM labman.plate libprepplate
+                LEFT JOIN labman.well libprepwell ON (
+                    libprepplate.plate_id = libprepwell.plate_id)
+                LEFT JOIN labman.composition libprepcpcp ON (
+                    libprepwell.container_id = libprepcpcp.container_id)
+                LEFT JOIN labman.library_prep_shotgun_process libpreppr ON (
+                    libprepcpcp.upstream_process_id = libpreppr.process_id)
+                LEFT JOIN labman.library_prep_shotgun_composition libprepcp ON
+                    (libprepcpcp.composition_id = libprepcp.composition_id)
+                LEFT JOIN labman.normalized_gdna_composition b ON (
```
AmandaBirmingham: Please rename the "b" and "c" aliases to normgdnacp and compgdnacp to match the naming conventions of this sql statement.

labcontrol/db/process.py
wasade: Ah okay

wasade: Perhaps update #506 with a mention of this?

labcontrol/db/process.py
```diff
@@ -3819,3 +3826,407 @@ def _generate_amplicon_prep_information(self):
...
+                LEFT JOIN labman.plate primersetplate2 ON (
+                    primersetwell2.plate_id = primersetplate2.plate_id)
+                FULL JOIN qiita.study_sample USING (sample_id)
+                LEFT JOIN qiita.study as study USING (study_id)
+                WHERE libprepplate.plate_id IN (
+                    SELECT distinct libprepplate2.plate_id
+                    FROM labman.sequencing_process sp
+                    LEFT JOIN labman.sequencing_process_lanes spl USING (
+                        sequencing_process_id)
+                    LEFT JOIN labman.pool_composition_components pcc1 ON (
```
AmandaBirmingham: In this query, unlike the one it was based on, the pool_composition_components table is only used once.  Therefore, please alias this as `pcc` rather than `pcc1` as we don't have to distinguish it from a second reference to the same table :)

labcontrol/db/process.py
charles-cowart: Short of changing the datatypes to a user-defined (domain) type, we could add constraints like these:
ALTER TABLE labman.well ADD CONSTRAINT rowchk CHECK (row_num > 0);
ALTER TABLE labman.well ADD CONSTRAINT colchk CHECK (col_num > 0);

labcontrol/db/process.py
```diff
@@ -3819,3 +3826,407 @@ def _generate_amplicon_prep_information(self):
...
+                """
+
+        with sql_connection.TRN as TRN:
+            TRN.add(sql, [self.id])
+
+            results = [dict(r) for r in TRN.execute_fetchindex()]
+
+            for d in results:
+                # assert that study_id and orig_name are only None when both
+                # are None; making this a control entry and not a sample entry
```
AmandaBirmingham: So, the problem is that this isn't necessarily true--at least not currently.  We definitely DO have real, experimental samples coming through and getting plated that happen NOT to be in qiita (which is all that these two pieces of being None indicates).  What is this field being used for?  Should we name it more correct, like "is_not_in_qiita" or something?  I wouldn't want to use it for anything that required that its contents really, truly be controls ...

labcontrol/db/process.py
charles-cowart: nope! nice catch

labcontrol/db/process.py
```diff
@@ -3819,3 +3826,407 @@ def _generate_amplicon_prep_information(self):
...
+                    # and/or 'orig_name' downstream, w/out losing important
+                    # metadata. Also clearly names this property.
+                    d['is_control'] = True
+                else:
+                    d['is_control'] = False
+
+                # convert time from datetime objects to strings
+                # no timezone information appears with %z; Assume timestamp
+                # w/out timezone type in Pg.
+                # therefore, assume timestamps are local to San Diego and are
```
AmandaBirmingham: I don't think this comment is relevant anymore now that Process.get_date_format is being used.

labcontrol/db/process.py
```diff
@@ -3819,3 +3826,407 @@ def _generate_amplicon_prep_information(self):
...
+                id = d['kappa_hyper_plus_kit_id']
+                d['kapa_hyper_plus_kit_lot'] = reagent[id]['external_lot_id']
+
+                id = d['stub_lot_id']
+                d['stub_lot_id'] = reagent[id]['external_lot_id']
+
+                # We have two robot IDs. Not sure which one is rightfully the
+                # 'extraction robot', but the example value is both strings
+                # separated by an underscore. Tentatively using this combined
+                # value.
```
AmandaBirmingham: Yes, that is what to use for now as BOTH of them are rightfully extraction robots--both are used in the extraction process.  Please remove this comment, but, if you want to you can reference https://github.com/jdereus/labman/issues/324  for the future :)

labcontrol/db/process.py
```diff
@@ -3819,3 +3826,407 @@ def _generate_amplicon_prep_information(self):
...
+                d['library_construction_protocol'] = 'KL KHP'
+
+                # EXPERIMENT_DESIGN_DESCRIPTION as with Amplicon, will remain
+                # empty when NULL.
+
+                # Replicating logic from Amplicon pre-processing
+                # TODO: refactor to a shared method
+                d['orig_name2'] = d['orig_name']
+
+                # Note is_control field replaces the line directly below
```
AmandaBirmingham: I don't understand this comment ?

labcontrol/db/support_files/patches/1.sql
```diff
@@ -1,3 +1,8 @@
 -- April 2, 2019
 -- Implicitly track the date and time of plate creation
 ALTER TABLE labman.plate ADD COLUMN creation_timestamp timestamp NOT NULL default now();
+
+-- May 9, 2019
+-- Add constraints to existing columns for wells.
+ALTER TABLE labman.well ADD CONSTRAINT rowchk CHECK (row_num > 0);
+ALTER TABLE labman.well ADD CONSTRAINT colchk CHECK (col_num > 0);
```
AmandaBirmingham: 👍

labcontrol/db/process.py
charles-cowart: Not a problem; although currently there is effectively only one prep_sheet_id, due to merging the data from the control and samples files.

labcontrol/db/tests/test_process.py
charles-cowart: Not sure where to go on this one; assuming the sample output file represents correct output, directly comparing the results of the method with established results seems effective.

The only thing I can think of is that the output of the SQL query isn't ordered, so it's perhaps possible for the row ordering to differ on an otherwise identical product.

wasade: Within this test, is there anything else that comes down in the call to `generate_prep_information`?

labcontrol/db/process.py
charles-cowart: It looks like the reason this line works, at least for the test data, is that the raw results from the SQL query can be organized into two groups. The first group is 288 in number, looks like sample data, and has no missing values.

The second group number 92 and uniformly have NULL values for the following five columns:
['orig_name', 'study_id', 'sample_id', 'project_name', 'experiment_design_description']

Hence, rows where both study_id and 'orig_name' are NULL matches all of the second set and none of the first.

It appears I arrived at this by negating the condition we use to test for a 'sample row' in Amplicon prep info file generation:
`if result['orig_name2'] is not None and study_id is not None:`
Note that orig_name2 is the same as orig_name at this point in the code, and both orig_name and study_id have not been modified since coming off the DB. However, I got it wrong; the negation should be if orig_name is None OR study_id is None.

With respect to the test data, it's amusing that both statements will work equally well in this particular case. IMHO, we could have an is_control column if determining which row represents a control is potentially fuzzy. I could also determine what is a control (or at least not a sample) by parsing the content column, and targeting rows with content starting with 'vibrio' or 'blank' or another word in the table.

I'll leave this one alone until @AmandaBirmingham comments, and then I can make the change easily enough.

labcontrol/db/tests/test_process.py
charles-cowart: Nothing. Generate_prep_information is just a fork to go to either amplicon prep info file generation or meta genomics prep info file generation, based on internal metadata. The only thing I can see coming back in addition to the output is the counterpart to exp_key itself, the 'obs_key' if you will.

charles-cowart:
@AmandaBirmingham Thanks! Will do!

labcontrol/db/process.py
wasade: Thanks, @charles-cowart

wasade:
I'm :+1: pending the reply / edit to this [comment](https://github.com/jdereus/labman/pull/502/commits/454855c6b46c0e104c134f23bad7cb31c0d87e99#r282764816)

labcontrol/db/process.py
AmandaBirmingham: Please 
(a) take the creation of the `is_control` field out.  It is not clearly named and thus has the potential to lead to gnarly bugs down the road when someone inaccurately assumes it is what it says it is (especially since our existing test conditions don't exercise the ways in which it secretly isn't what it says it is).  Since it is currently used in only one place, we have the option to just drop it until we come up with a better naming/handling of this weird "maybe controls but maybe just unloved experimental samples" situation.
(b) replace the current use of `is_control` to decide when to modify `orig_name2` with the check we use for that in `generate_amplicon_prep_info`:

`if result['orig_name2'] is not None and study_id is not None:`

As I recall, it took you and me about a week of back-and-forth to get this condition *right*, and so I don't want to mess with it again (at least not till we do a fuller refactor someday).

charles-cowart: Too true!  😆 

Fair enough, although it seems like we could really benefit from easily identifying states. It is something I hope we can get back to at some point.

charles-cowart:
@AmandaBirmingham @wasade Thanks for all your comments! I put the most recent request changes in, and it's currently going through Travis...

AmandaBirmingham:
This looks good to me! @charles-cowart  can I just confirm that you've run LabControl locally and verified that you can now see and successfully use the "Download Preparation Sheets" button for the existing shotgun sequencing run in the test database?

charles-cowart:
Thanks @AmandaBirmingham ! Yup, downloads as expected. :)

PR #515, opened 2019-05-09
Issue438

### LINKED ISSUE #438: Change labman/labman/db/process.py generate_echo_picklist() call to _format_picklist() to pass in 20000 (nL) as max_vol_per_well
From RodolfoSalido:

I quantified my Shotgun libraries and generated a pooling file using Lab Control. I only noticed one discrepancy between the pooling file generated by Lab Control and that generated by our Jupyter notebook. Our Jupyter notebook is designed to pool into a destination well until it reaches a certain volume (30µL) and then it cycles to the next well. The Lab Control pooling function seems to have a value of 60µL for the max_vol_per_well parameter instead of 30µL.

### LINKED ISSUE #505: PoolingProcess._format_picklist() produces incorrect output if populating >15 destination wells (on a 384-well plate)
For the test data for PoolingProcess(3) in the test db, the volume in every well of the 384-well input plate is 1 nL except for wells 0 23 and 24 and P 23 and 24, which are empty.  In this, case, when entering a max_vol_per_well of 1 nL, one would expect each input well to go into a separate output well (except for the 4 noted above), so there should be 380 output wells (A1-P20).  See attached expected output:

[expected_poollist.txt](https://github.com/jdereus/labman/files/3160400/expected_poollist.txt)

However, the code in fact produces output that works as expected up to A15 and then sends output to B16-B23, then B0[sic]-B7, then C8-C23, and so on in unexpected ways until X20 (spoiler: not a real 384-well-plate well :) See attached buggy output:

[buggy_poollist.txt](https://github.com/jdereus/labman/files/3160408/buggy_poollist.txt)

AmandaBirmingham:
- Fixes #438 : see changes to `generate_echo_picklist` and addition of `test_generate_echo_picklist_nondefault_volume` with supporting new test data file `low-max-vol-pooling-process-echo-picklist.txt`.
- Fixes #505 : see refactor of `_format_picklist` and rewrite of `test_format_picklist` as well as addition of `test_format_picklist_error_dest_plate_too_small` and `test_format_picklist_error_input_vol_too_large`.
- Addresses small flake8 issues in `test__folder_scrub_name`.


labcontrol/db/process.py
```diff
@@ -2520,35 +2520,75 @@ def _format_picklist(vol_sample, max_vol_per_well=60000,
...
-                well_name = "%s%d" % (chr(ord('A') + i), j + 1)
-                # Machine will round, so just give it enough info to do the
-                # correct rounding.
-                val = "%.2f" % pool_vols[i][j]
-                # test to see if we will exceed total vol per well
-                if running_tot + pool_vols[i][j] > max_vol_per_well:
-                    d += 1
-                    running_tot = pool_vols[i][j]
+
+        for curr_input_row_index in range(num_input_rows):
```
wasade: I believe that weirdness would result if `num_input_rows` is `>= 26`. Similar to the [comment](https://github.com/jdereus/labman/pull/502/files#r282702728) in #502, should there be defensive bounds checking (...or is it somewhere deeper in the codebase...)?

labcontrol/db/process.py
```diff
@@ -2520,35 +2520,75 @@ def _format_picklist(vol_sample, max_vol_per_well=60000,
...
+            for curr_input_col_index in range(num_input_cols):
+                curr_input_well_name = get_well_name(curr_input_row_index,
+                                                     curr_input_col_index + 1)
+
+                # test to see if adding the volume of this input well to the
```
wasade: It looks like the addition is performed on line 2560?

Is this comment in reference to the `if` on line 2554? I think I'm getting tossed up by the proximity of the comment. Would it be possible to modify this comment to something along the lines of "test to see if the current well exceeds the total volume per well. If it does not, then test to see if adding requires incrementing the well index"? 

~~I almost wonder if the check performed on line 2554 would make sense to perform in bulk at or near line 2542?~~ Scratch that, that would complicate providing a user digestible error message.

labcontrol/db/process.py
AmandaBirmingham: Oooo, great point--I like the cut of your edge-case jib!  However, we should be careful about where we add this check because there is nothing that inherently prevents a plate from having >26 rows (e.g., 1584-well plates have 32 rows) or that prevents the majority of this code from handling it.  The specific issue would, I believe, only be that the (newly encapsulated :D ) code here for generating the well NAMES wouldn't work with >26 rows and would need to be refactored accordingly (e.g., 1584s roll over to AA, AB, etc, like excel columns). I will therefore add a check in that sub-method that throws a ValueError if the offset is >25.  

That said, there is code to either generate or parse well names in various other places in LabControl too -- e.g., in the generate_prep_information methods where you caught this, and at least in the two additional locations below, and maybe more:

https://github.com/jdereus/labman/blob/645c798e0094e9962973c2d111c4860b1e8e5b23/labcontrol/db/container.py#L269-L279

https://github.com/jdereus/labman/blob/7bb9106555143bb4877c49516e92c792ea9349a5/labcontrol/db/process.py#L1828-L1861

Based on this, I suggest we also make an issue to do a sweep for affected code, because it definitely isn't all centralized ...

labcontrol/db/process.py
AmandaBirmingham: Yeah, I see what you mean.  I will move/tweak the comment.

labcontrol/db/process.py
wasade: :+1:. Excellent point with the >26 case and the other instances. Agree that a sweep issue makes sense -- thank you for the additional investigative work!

labcontrol/db/process.py
wasade: Thanks!!

AmandaBirmingham:
@wasade The new additions to this PR have passed build :) Could you take a look at them and see if you approve?

PR #516, opened 2019-05-10
BLD: adding versioneer

### LINKED ISSUE #487: Describe versioning scheme
For milestone 4, we began associating versions as at the completion of the milestone, a new version of labcontrol will be cut.

The repo presently, and loosely, uses the QIIME2 syntax, e.g. 2019.4. For this milestone, it is currently labeled using a more traditional semantic versioning, e.g., 0.1.0. Both versioning schemes are in use with projects in the lab. The 0.1.0 format as major version bump indicates things such as full launch, minor version bumps being functionality, major bug fixes, etc, and microversion bumps being no change to the overall user experience but project improvements such as the test framework. 

However, there is certainly merit for versioning based off of date as it is immediately clear when the release was. 

If there is a strong preference for one over the other, then let's proceed with the preference. The label if this milestone is not set in stone and will be updated if necessary for use in the associated release.

wasade:
Fixes #487.

This is using common `versioneer` configuration with QIIME2, and a version string that confirms to [PEP440](https://www.python.org/dev/peps/pep-0440/). There are a few pleasant things about `versioneer`, first being that we never have to specify versions only tags! A tag name is implicitly added into the version, so once the repository is tagged for release (e.g., `2019.5.0`), it will be automagically incorporated into the version string, as will date and commit hash.


charles-cowart:
I wonder if there is a way to have these files ignored, w/respect to coverage.

wasade:
@AmandaBirmingham, comes from the [install doc](https://github.com/jdereus/labman/pull/516/files#diff-f374b985f7c07f7ff6a8dbaa0b30a10dR28) which is also in the project readme. I'm guessing the maintainer fell on this due to the chicken and egg problem when the package isn't installed as it povides an out of band means to obtain version information. Happy to explore further if you would like.

wasade:
@charles-cowart good question! I bet there is, will explore momentarily

AmandaBirmingham:
@wasade  Nope nope, no need to explore further :)

Issue #522, opened 2019-05-21
Implement tester.py equivalent for gui


AmandaBirmingham:
populate w code that loads index.html (the test runner page) into headless chrome and captures results

charles-cowart:
@josenavas, @tanaes Do you guys remember why we have tester.py? It appears to be something to load test classes and run them according to supplied parameters. However, all the tests are subclassed from LabControlTestCase, which in turn is subclassed from unit test's own TestCase class. Hence, I'm kind of wondering why we have a second entry point to run unit-tests? It seems like we could run every test through nosetests?

charles-cowart:
Related to #523. See comments.

charles-cowart:
Resolved by #527

Issue #523, opened 2019-05-21
Integrate gui tester.py into .travis.yml


AmandaBirmingham:

charles-cowart:
Summary of how the javascript tests are run in Emperor:

A line in the Travis YAML file executes 'coverage run tests/all_tests.py'; Coverage is simply wrapping the execution of all_tests.py, a script with a command-line interface that can run some or all of Emperor's tests, based on parameters.

PhantomJS is a headless browser based on WebKit. A section of code in all_tests.py builds a command-line argument to execute phantomjs as console program. The argument takes runner.js and index.html as parameters.

Runner.js is a stock support script for running Qunit tests using PhantomJS as the execution engine.  Index.html declares all of the tests that need to be performed. As the tests are run through PhantomJS, information outputs to stderr and stdout. If all tests pass, the exit code for the PhantomJS process will be 0, otherwise, it will be a non-zero value.

The all_tests.py code captures the stderr, stdout and return code from the exec() of PhantomJS, potentially displays stderr and stdout, and processes the return code into a success or failure boolean in the Qunit framework.

charles-cowart:
With the above confirmed, and headless chrome a Travis dependency, a toy version of this workflow will be forthcoming.

wasade:
Thanks, @charles-cowart!

cc @ElDeveloper (just to keep appraised, no specific request)

charles-cowart:
@AmandaBirmingham My pleasure!

Just an FYI to anyone interested, it seems as though the latest auto-update to Chrome causes an error when trying to run headless. Currently, I can dismiss two of the errors by specifying my own --crash_dir, but I have not been able to resolve the remaining error:

[0524/104021.267190:ERROR:browser_process_sub_thread.cc(217)] Waited 1 ms for network service

This appears to occur for others as recently as May, It doesn't appear to be resolved yet. In the meantime, I've found a source for the latest v73 build of Chromium, which doesn't have this issue:

https://commondatastorage.googleapis.com/chromium-browser-snapshots/index.html?prefix=Mac/625897/

charles-cowart:
Committing my notes to GitHub:

Testing in LabControl can be broken down into two parts: The unit-tests that are subclassed from unittest and run through nosetests, and the stress and integration tests, which are executed using db/tester.py.

[...]

PR #527, opened 2019-05-24
Test npm, etc. installation

charles-cowart:
Test installation of npm, qunit, headless-chrome, and the interface script between the last two on a trivial assertion test.


charles-cowart:
Command "node-qunit-puppeteer labcontrol/gui/ctests/index.html 10000  "--allow-file-access-from-files --no-sandbox"" works as intended locally, but is failing with 'TypeError: console.group is not a function' in Travis. Will confirm versions both locally and in Travis tonight as a next step.

charles-cowart:
I believe @AmandaBirmingham wishes to merge this request before merging her own, hence the earlier it can be reviewed and merged, the better.

.travis.yml
```diff
@@ -20,6 +23,12 @@ before_install:
   - export PATH=/home/travis/miniconda3/bin:$PATH
   # Update conda itself
   - conda update --yes conda
+  # test that npm exists with language: node_js specified.
+  # node-qunit-puppeteer is a runner.js like interface for
```
AmandaBirmingham: I do not understand the comment "... exists with language: node_js specified" ... I don't see node_js specified ... could you clarify?

.travis.yml
```diff
@@ -20,6 +23,12 @@ before_install:
...
+  # headless-chrome and qunit.
+  # node-qunit-puppeteer installs its own chromium as a dependency.
+  # test forgoing of chrome as an addon.
+  - npm install -g node-qunit-puppeteer
```
AmandaBirmingham: The comment "test forgoing of chrome as an addon" is also opaque to me; please clarify.  
Also, "foregoing" not "forgoing" :D

.travis.yml
```diff
@@ -56,15 +65,17 @@ install:
   # Deactivate the Qiita environment
   - conda deactivate
   # Set the labcontrol config file
-  - sed "s#CERTIFICATE_FILEPATH=#CERTIFICATE_FILEPATH=$PWD/support_files/server.crt#g" labcontrol/db/support_files/test_config.cfg > ~/.labcontrol.cfg.1
+  - sed "s#CERTIFICATE_FILEPATH=#CERTIFICATE_FILEPATH=$PWD/support_files/server.crt#g" labcontrol/db/support_files/qiita_test_config.cfg > ~/.labcontrol.cfg.1
```
AmandaBirmingham: Why was this file renamed to "qiita_test_config".cfg?  It holds considerably more than qiita config info ...

.travis.yml
```diff
@@ -76,7 +87,12 @@ install:
   - psql -d qiita_test -c 'select * from qiita.study;'
   - labcontrol patch
 script:
-  - nosetests --with-doctest --with-coverage -v --cover-package=labcontrol
+  # test javascript unit-testing using qunit and chrom{e|ium} using the
+  # node-qunit-puppeteer script and a trivial qunit-based test in a new
+  # directory called ctests.
+  # - node-qunit-puppeteer labcontrol/gui/test/index.html 10000  "--allow-file-access-from-files --no-sandbox"
```
AmandaBirmingham: I think this comment, and the commented-out line, are somewhat out-of-date.  Could you update?

labcontrol/gui/js_tests/index.html
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
```
AmandaBirmingham: I do not understand why we are checking in this file ... isn't it a POC test that we don't want in the master codebase?  Or does the testing fail if it has nothing to work on?

labcontrol/gui/js_tests/js_tester.py
```diff
@@ -0,0 +1,130 @@
+import click
+from os.path import join, abspath, dirname
+from subprocess import Popen, PIPE
+import os
+
+
+'''
+Notes: It's generally not a good idea to have Click commands call other click
+commands, although ti can be done using context objects.
```
AmandaBirmingham: typo: "ti" for "it"

labcontrol/gui/js_tests/js_tester.py
```diff
@@ -0,0 +1,130 @@
...
+# note, the default has been explicitly defined here as 30,000ms. If none is
+# supplied, the node_qunit_puppeteer script itself will default to 30,000ms.
+@cli.command()
+@click.option('--timeout_in_milliseconds', required=False,
+              type=click.IntRange(0, None),
+              default=30000, show_default=True,
+              help='timeout for a unittest, in milliseconds')
+def all_tests(timeout_in_milliseconds):
+    '''Command to run all tests managed by this script. Currently there is only
+    one (javascript qunit tests).
```
AmandaBirmingham: Guidance for 1-line docstring: "It prescribes the function or method's effect as a command ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ..."." (PEP 257) ... can this be revised into this format?

Also, why "javascript qunit test"?  ... I would have expected this comment would reference "run_js_test"

labcontrol/gui/js_tests/js_tester.py
```diff
@@ -0,0 +1,130 @@
...
+    # html files that are located within the same directory as this file
+    test_dir = abspath(dirname(__file__))
+
+    # if a single test fails, this value will become False
+    all_tests_successful = True
+
+    # build a list of html files to process. The contents of this list will
+    # then be processed in turn. The reason for this is to discover the number
+    # of files to be processed, and set up the progress bar.
+    html_files = []
```
AmandaBirmingham: The comments in this function are awesome--really helpful!

labcontrol/gui/js_tests/tests.js
```diff
@@ -0,0 +1,7 @@
+QUnit.test( "hello test", function( assert ) {
```
AmandaBirmingham: Same comment here as for index.html

labcontrol/gui/js_tests/js_tester.py
```diff
@@ -0,0 +1,130 @@
...
+
+    for path, sub_directories, files in os.walk(test_dir, topdown=False):
+        for file in files:
+            if file.endswith('.html'):
+                html_files.append(join(path, file))
+
+    click.echo("%d html files found" % len(html_files))
+
+    count = len(html_files)
```
wasade: Minor, but maybe shift this to happen prior to line 98, and then use `count` in the string formatting?

labcontrol/gui/js_tests/js_tester.py
```diff
@@ -0,0 +1,130 @@
...
+
+    # similarly, store the stdout of each test in results, instead of
+    # displaying them as they are returned to us. This allows the progress bar
+    # to properly stay in a single location and reach 100% when all tests have
+    # completed.
+    results = []
+
+    with click.progressbar(length=count) as bar:
+        for html_file in html_files:
+            status, stdout = run_js_test(timeout_in_milliseconds, html_file)
```
wasade: Can `status` be renamed as `success` which I think is more accurate to its value? 

It may be pleasant to have `stdout` renamed to `test_output`. The motivation is that `stdout` does refer to a specific importable file descriptor (i.e., `sys.stdout`). This change is just defensive to avoid possible weird bugs in the future :)

.travis.yml
charles-cowart: No problem. I realize now this is a comment for me; I wasn't sure if 'npm' would be available by adding 'language: node_js' up top or not. It's no longer relevant, so I'll remove it.

.travis.yml
charles-cowart: Another 'note-taking' comment made during development - I'll replace with, "hence, a separate chrome installation is not required." I was installing chrome as an individual package. I removed it when I discovered node-qunit-puppeteer installs its own chromium.

Sorry, ran out of 'e's on Friday. :D

.travis.yml
charles-cowart: My bad. You're right, this is a labcontrol config file; I mistook it for a sample Qiita config file. Something other than 'test_config.cfg' might be more appropriate, as it can be interpreted as configuration information for Labcontrol's own tests; what do you think? Perhaps 'sample.cfg' or 'travis_build.cfg' might be more illuminating.

.travis.yml
charles-cowart: Done, thanks for the catch!

labcontrol/gui/js_tests/js_tester.py
charles-cowart: Thanks!

labcontrol/gui/js_tests/js_tester.py
charles-cowart: Agreed, yeah that is rather dumb. :)

labcontrol/gui/js_tests/index.html
charles-cowart: It's just there as a reference; we'd replace it with your tests. Good point though, I should verify its behavior with zero tests.

charles-cowart: Verified behavior is currently script will return 0 (success) when zero tests are found.

.travis.yml
AmandaBirmingham: It isn't just used for tests.  I don't think I named it, but I suspect it was named "test_config" not because it is related to unit tests, but because it is the exercises (tests) the minimal config settings that one needs to run the software (back before the certificate stuff came in, this file could work as the config without any user modification).  

Now it is sort of a skeleton config, not really a working config on its own.  We could call it labcontrol_config.cfg, if you like, or skeleton_labcontrol_config.cfg, or base_labcontrol_config.cfg, or ....?

wasade: :+1: on latter, "travis_build.cfg"

labcontrol/gui/js_tests/tests.js
charles-cowart: I suppose my thinking was to submit these files so they're in the repo and we have a complete, if trivial working framework. I can drop them from the PR however.

.travis.yml
AmandaBirmingham: We *could* call it "sample_config.cfg" (per your suggestion) but I'm vaguely uncomfortable with that because, nowadays, it isn't a working config file right out of the box--the user has to fill in some more values.  Let's definitely stay away from "travis_build.cfg" because this config file is used when the end user sets up the software, not just during travis building.

.travis.yml
AmandaBirmingham: > ran out of 'e's on Friday

 Bahahahahaha! Excellent :)

labcontrol/gui/js_tests/tests.js
AmandaBirmingham: Ok, I guess we can leave them in ... but when I put in my PR that involves some real LabControl tests, I assume I'm taking them right back out, yes?

.travis.yml
charles-cowart: 'cfg' itself implies config - maybe removing '_config' might make these a little less wordy:
sample.cfg
labcontrol.cfg
skeleton_labcontrol.cfg
base_labcontrol.cfg

In some of the open-source software I use, it's typical to have these incomplete config files as 'something.conf.sample' - perhaps these might also be a good alternative:

labcontrol.conf.sample
labcontrol.cfg.sample

labcontrol/gui/js_tests/js_tester.py
charles-cowart: Will do!

.travis.yml
AmandaBirmingham: I vote for skeleton_labcontrol.cfg or base_labcontrol.cfg and against the use of the name "sample", which is used all over labcontrol to mean something different than it would mean here (and what it would mean here isn't what I would expect it to mean :)

charles-cowart: I vote for 'skeleton_labcontrol.cfg' over 'base_labcontrol.cfg', as it better implies that this file is incomplete.

AmandaBirmingham: Sold!

charles-cowart:
Resolves #522, #523

PR #548, opened 2019-07-06
Kappa fix + initial logging support

charles-cowart:
Work in progress - testing fix for 'Kappa Hyper Plus kit' to 'Kapa Hyper Plus kit', etc. without presently considering the patch system.


charles-cowart:
Not part of the bug fix itself, but I took the liberty of inserting a root logging.logger in the main thread, similar to the logging module we used for debugging qiita_client and plugins.

charles-cowart:
@fedarko @qiyunzhu I've tested the renaming process against an existing database as well as fresh databases, and the 2.py patch works as intended. I don't think it's necessary to understand the entire codebase to review this PR; reviewing the logging and the string changes should be fine, thanks!

docs/index.rst
```diff
@@ -259,7 +259,7 @@ This brings you to the “Shotgun library prep” page. Here, you will choose th
 
 This will bring up a text box asking for information on your shotgun libraries. Here you will be asked to select index primers by separately selecting your “i5 plate” and your “i7 plate”. As with equipment like robots, primer plates are expected to be added infrequently, and must already exist in the database.
 
-You will then be asked to name your plate, input your “kappa hyper plus kit”, input your “Stub lot” and input your “Volume (mL)”. Though all of the options are free text, if the “kappa hyper plus kit” and “Stub lot” do not exist, you will be prompted to add them to the system.
+You will then be asked to name your plate, input your “kapa hyper plus kit”, input your “Stub lot” and input your “Volume (mL)”. Though all of the options are free text, if the “kapa hyper plus kit” and “Stub lot” do not exist, you will be prompted to add them to the system.
```
qiyunzhu: Just for information, the original spelling seems to be [KAPA HyperPlus Kit](https://sequencing.roche.com/en/products-solutions/by-category/library-preparation/dna-library-preparation/kapa-hyperplus.html)

labcontrol/db/process.py
```diff
@@ -4032,9 +4032,8 @@ def _get_metagenomics_data_for_prep(self):
                 # query.
                 d['instrument_model'] = inst_mdl
 
-                # note that the correct term is 'Kapa', not 'kappa'.
-                id = d['kappa_hyper_plus_kit_id']
-                d['kappa_hyper_plus_kit_lot'] = reagent[id]['external_lot_id']
+                id = d['kapa_hyper_plus_kit_id']
```
qiyunzhu: I had the impression that `id` is a reserved Python keyword. So I usually use `id_`.

labcontrol/db/support_files/db_patch.sql
```diff
@@ -477,15 +477,15 @@ CREATE INDEX idx_16s_library_prep_composition_1 ON labcontrol.library_prep_16s_c
 CREATE TABLE labcontrol.library_prep_shotgun_process (
 	library_prep_shotgun_process_id bigserial  NOT NULL,
 	process_id           integer  NOT NULL,
```
qiyunzhu: I recall once in labcontrol they used `bigint` instead of `integer`. Is there a consideration for this change?

labcontrol/db/support_files/patches/python_patches/2.py
```diff
@@ -0,0 +1,90 @@
+from labcontrol.db.sql_connection import TRN
+import logging
+
+
+with TRN:
```
qiyunzhu: If I understand correctly, these codes are to resolve some historical issue of naming inconsistency. Is that right?

labcontrol/db/support_files/patches/python_patches/2.py
```diff
@@ -0,0 +1,90 @@
...
+    # rename fk_shotgun_library_prep_process_kappa in
+    # library_prep_shotgun_process
+    #
+    sql = """SELECT COUNT(*) FROM information_schema.constraint_column_usage
+                 WHERE table_name = 'library_prep_shotgun_process' AND
+                 constraint_name = 'fk_shotgun_library_prep_process_kappa';"""
+
+    logging.debug(sql)
+    TRN.add(sql)
+    result = TRN.execute_fetchflatten()[0]
```
qiyunzhu: I wonder if `fetchflatten` has too strong assumption of the data structure. Maybe it's good to comment that the first column is always something.

scripts/labcontrol
```diff
@@ -15,6 +15,43 @@ import click
...
+# logging level defaults to CRITICAL, the least verbose option.
+if 'LABCONTROL_DEBUG_LEVEL' in environ:
+    level = environ['LABCONTROL_DEBUG_LEVEL']
+    if level in debug_levels_list:
+        logging_level = debug_levels_list[level]
+    else:
+        raise ValueError(
+            "%s is not a valid value for LABCONTROL_DEBUG_LEVEL" % level)
+
+# a software project may define multiple loggers, for instance in different
```
qiyunzhu: Thanks for the explanation!

scripts/labcontrol
```diff
@@ -15,6 +15,43 @@ import click
...
+# the default logging level
+logging_level = logging.CRITICAL
+
+# the user can change the root logger's level using the LABCONTROL_DEBUG_LEVEL
+# environment variable. It must be set to a valid string in debug_levels_list,
+# or else an error will be raised. If this environment level is not set, then
+# logging level defaults to CRITICAL, the least verbose option.
+if 'LABCONTROL_DEBUG_LEVEL' in environ:
+    level = environ['LABCONTROL_DEBUG_LEVEL']
+    if level in debug_levels_list:
```
qiyunzhu: How about the `try ... except` structure, which should be cheaper.

labcontrol/db/process.py
charles-cowart: To my knowledge, 'id' is not reserved. I know that 'if', 'is', and 'in' are.

labcontrol/db/support_files/db_patch.sql
charles-cowart: Thanks for asking! I believe the standard integer in PostgreSQL 9 is faster than bigint, and still four bytes long - that would make for over 2 billion process IDs, even if we ignore signage. We're probably pretty good in this case.

labcontrol/db/support_files/patches/python_patches/2.py
charles-cowart: right. There was some confusion between the proper name, and the Greek letter kappa.

labcontrol/db/support_files/patches/python_patches/2.py
charles-cowart: Thanks! Quite used to it from Qiita, but yeah it doesn't read well. I'll add a comment.

scripts/labcontrol
charles-cowart: This is more explicit, IMHO. Not sure if it would be cheaper than a conditional, either.

qiyunzhu: I googled a bit and found diverse opinions. Here's the official explanation: [EAFP](https://docs.python.org/3/glossary.html#term-eafp). Totally trivial. Never mind:)

docs/index.rst
charles-cowart: Thanks! Changes made.

scripts/labcontrol
charles-cowart: I do like it better aesthetically...

charles-cowart:
@qiyunzhu Thanks Qiyun! Would you mind looking over the changes one last time? If you approve, can you please merge the code? Thanks!

qiyunzhu:
@charles-cowart Good job! Merging.

PR #556, opened 2019-08-06
Case sensitivity fix

charles-cowart:
Resolves issue in plate-mapper where the pull-down menu doesn't populate correctly if the search term is correct, but in a different case (upper vs lower vs mixed, etc.).


charles-cowart:
closing. Incorporated changes from PR 554 as well by accident.

PR #557, opened 2019-08-06
Case-insensitive pull-down menu in plate-mapper

charles-cowart:
Resolves issue in plate-mapper where the pull-down menu doesn't populate correctly if the search term is correct, but in a different case (upper vs lower vs mixed, etc.).

### DIFF
--- a/labcontrol/db/study.py
+++ b/labcontrol/db/study.py
@@ -10,6 +10,8 @@ from . import base
 from . import sql_connection
 from . import user
 
+import logging
+
 
 class Study(base.LabControlObject):
     """Study object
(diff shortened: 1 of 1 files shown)


qiyunzhu:
I read the code but I am not able to understand though...

charles-cowart:
> I read the code but I am not able to understand though...

@qiyunzhu That's okay, just verify the changes for general correctness. Also, although the check is failing, you can see that it's failing on the prep information file test I wrote a fix for in another PR, so it's fine to approve this patch based on your approval of the code changes.

qiyunzhu:
@charles-cowart Thanks for explanation! Code looks fine to me--at least literally. Will approve.

PR #566, opened 2019-08-17
Modified PlateViewer colors + Legend

charles-cowart:
Text for invalid specimen ids implied id would not exist across all
studies, when it's only against the studies associated with the plate.

Added a new yellow candy-cane/barber-bar to represent a cell that was
populated using a multiselect paste and has more than one possible
specimen ID match across the attached studies. This is so the user can
be alerted, click on the cell, and select from the drop-down that should
appear.

Modified the colors to be more appealing. Looking for collective
feedback. Bold red is gone, and an invalid specimen ID is now the red
candy-cane color. Blue candy-cane color is the same as before,
representing a specimen ID that has already been plated on another
plate.

A new Yellow/Orange candy-cane stripe identifies values that match more
than one possible specimen ID from the attached studies.

The candy-cane types are mutually exclusive; a cell can't be both valid
and invalid (0 possible matches), invalid and matching 2 or more
possible specimen ids, etc.

However, any of the above types can still have duplicate entries on the
current plate. Hence, the platemapper now identifies duplicate entries
by wrapping the cell in a purple border. This way, there doesn't have to
be a hierarchy defining which error should be presented over another,
and it won't be random based on implementation.

Lastly, as the purple border can overwrite the green border for a cell
with comment, cells with comments will now have green lettering, rather
than a green border.

new_legend

### DIFF
--- a/labcontrol/gui/static/css/labcontrol.css
+++ b/labcontrol/gui/static/css/labcontrol.css
@@ -24,14 +24,18 @@ cursor: pointer;
 background-color: white;
 }
 
+#plate-map-legend-commented-box {
+    text-align: center;
+}
+
 /* Wells class */
 
-/* Commented cell (green ringed cell) */
+/* Commented cell (green text) */
 .well-commented {
     color: green;
 }
 
-/* Value in cell duplicated on current plate (yellow ringed cell) */
+/* Value in cell duplicated on current plate (purple ringed cell) */
 .well-duplicated {
     border: 2px solid purple;
 }
--- a/labcontrol/gui/static/css/labcontrol.css
+++ b/labcontrol/gui/static/css/labcontrol.css
@@ -25,20 +25,18 @@ background-color: white;
 }
 
 /* Wells class */
+
+/* Commented cell (green ringed cell) */
 .well-commented {
-    border: 2px solid green;
+    color: green;
 }
 
+/* Value in cell duplicated on current plate (yellow ringed cell) */
 .well-duplicated {
-    background: repeating-linear-gradient(
-        45deg,
-        #f2a4a4,
-        #f2a4a4 10px,
-        #edc9c9 10px,
-        #edc9c9 20px
-        ) !important;
+    border: 2px solid purple;
 }
 
+/* Value in cell has already been plated on another plate */
 .well-prev-plated {
     background: repeating-linear-gradient(
         45deg,
@@ -49,9 +47,26 @@ background-color: white;
         );
 }
 
+/* Value in cell is not a member of the attached studies */
 .well-unknown {
-    background: #ff0000;
-    color: white;
+    background: repeating-linear-gradient(
+        45deg,
+        #f2a4a4,
+        #f2a4a4 10px,
+        #edc9c9 10px,
+        #edc9c9 20px
+        );
+}
+
+/* Value in cell matches one or more members of the attached studies */
+.well-indeterminate {
+    background: repeating-linear-gradient(
+        45deg,
+        #f2f2a4,
+        #f2f2a4 10px,
+        #FFE4B5 10px,
+        #FFE4B5 20px
+        );
 }
 
 .comment-box {
--- a/labcontrol/gui/templates/plate.html
+++ b/labcontrol/gui/templates/plate.html
@@ -159,31 +159,46 @@
 
     <div id='plate-map-div' class='spreadsheet-container'></div>
 
-    <!-- Plate map legend div -->
+    <!-- Plate map legend div
+        Note that well-unknown and well-prev-plated are mutually exclusive
+        while well-duplicated is not. A duplicated value can appear in the
+        plate and be valid, unknown, or previously plated on another plate. -->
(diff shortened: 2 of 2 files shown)


labcontrol/gui/static/css/labcontrol.css
```diff
@@ -25,20 +25,18 @@ background-color: white;
 }
 
 /* Wells class */
+
+/* Commented cell (green ringed cell) */
```
fedarko: The green isn't applied to the border any more, so this should say something like "cell with green text" instead of "green ringed cell".

labcontrol/gui/static/css/labcontrol.css
```diff
@@ -25,20 +25,18 @@ background-color: white;
...
 .well-commented {
-    border: 2px solid green;
+    color: green;
 }
 
+/* Value in cell duplicated on current plate (yellow ringed cell) */
```
fedarko: Should say purple instead of yellow.

labcontrol/gui/templates/plate.html
```diff
@@ -159,31 +159,46 @@ <h4>Plate map</h4>
 
     <div id='plate-map-div' class='spreadsheet-container'></div>
 
-    <!-- Plate map legend div -->
+    <!-- Plate map legend div
+        Note that well-unknown and well-prev-plated are mutually exclusive
+        while well-duplicated is not. A duplicated value can appear in the
```
fedarko: minor but the wording of this comment is a bit unclear, it's initially confusing as to what `well-duplicated` "is not".

something like "Although `well-prev-plated` is mutually exclusive with `well-unknown`, `well-duplicated` is not." would make more sense IMO

labcontrol/gui/templates/plate.html
```diff
@@ -159,31 +159,46 @@ <h4>Plate map</h4>
...
+        plate and be valid, unknown, or previously plated on another plate. -->
     <div id='plate-map-legend'><h5>Plate map legend</h5>
         <div id='legend-rows' class='legend-container'>
           <section id='plate-map-legend-commented' class='legend-entry'>
-            <div id='plate-map-legend-commented-box' class='well-commented legend-patch'></div>
+            <div id='plate-map-legend-commented-box' class='well-commented legend-patch'><center>Specimen ID</center></div>
```
fedarko: I know that this text being centered isn't critical, but the `<center>` tag is [deprecated](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/center)—we can't guarantee it'll be supported by modern browsers. A better solution would just be adding a CSS rule that says

```css
#plate-map-legend-commented-box {
    text-align: center;
}
```

...or something like that.

labcontrol/gui/templates/plate.html
```diff
@@ -159,31 +159,46 @@ <h4>Plate map</h4>
...
+              <p> Specimen ID is not a member of the attached studies</p>
+            </div>
+          </section>
+          <section id='plate-map-legend-indeterminate' class='legend-entry'>
+            <div id='plate-map-legend-indeterminate-box' class='well-indeterminate legend-patch'></div>
+            <div id='plate-map-legend-indeterminate-text' class='legend-text'>
+             <p> Specimen ID matches more than one member of the attached studies</p>
+            </div>
+          </section>
+          <section id='plate-map-legend-indeterminate' class='legend-entry' style="visibility: hidden">
```
fedarko: I'm unclear on why there's an invisible `<section>` here. I'm guessing this is a hack to make the layout work properly? (I have no room to judge on those sorts of things, don't worry :)

At the very least, it would be a good idea to document *why* this invisible `<section>` exists so that future developers aren't confused.

gwarmstrong: I'm not at my computer right now so I can't check, but I would imagine this is to make the legend centered on the last row. @charles-cowart if confirmed, I would reccomend something like:
```
<!-- Include the section below if there are an odd number of legend entries -->
<section id='plate-map-legend-for-spacing class='legend-entry' style="visibility: hidden">
  <div id='plate-map-legend-for-spacing' class='legend-patch'></div>
  <div id='plate-map-legend-for-spacing' class='legend-text'></div>
</section>
```
There might be a more minimal solution.

labcontrol/gui/static/css/labcontrol.css
charles-cowart: Will do!

labcontrol/gui/templates/plate.html
charles-cowart: Ha! I knew you would catch that! Thanks for the replacement!

labcontrol/gui/templates/plate.html
charles-cowart: I'm going to document it as is - basically the hidden section is there because otherwise the 'Well comments' header is offset to the middle. I'll leave it to you guys to decide if it's worth replacing in a subsequent PR.

fedarko: Sounds good to me. As long as the code is documented, I don't think replacing this in the future is important.

Issue #562, opened 2019-08-14
No plates show up if refreshing the "Plates list" page


fedarko:
When you first load the "Plates list" page, the `Plate type` dropdown is set to `Choose plate type...`

<img width="1552" alt="Screen Shot 2019-08-14 at 1 32 36 PM" src="https://user-images.githubusercontent.com/4177727/63054095-0f268580-be98-11e9-957f-d50066b6611c.png">

When the user selects a plate type, the associated plates are loaded as expected (...pardon the debug names).

<img width="1552" alt="Screen Shot 2019-08-14 at 1 38 51 PM" src="https://user-images.githubusercontent.com/4177727/63054530-e05cdf00-be98-11e9-96eb-2d7a7cd33c79.png">

However, if you navigate to another page (say "Plate search") while these plates are showing up, and then press your browser's back button to return to the "Plates list" page, you end up in a state where no plates are shown even though they currently exist for the selected (in this case, `Sample plates`) plate type:

<img width="1552" alt="Screen Shot 2019-08-14 at 1 35 55 PM" src="https://user-images.githubusercontent.com/4177727/63054294-73e1e000-be98-11e9-9219-b96516286ba7.png">

I've run into this sort of error with other web applications (and with AG Web, see [this issue](https://github.com/biocore/american-gut-web/pull/720) for an example). The way I see things, the easiest solution here is just ensuring that (on the page being loaded, regardless of whether it was from the user pressing the back button or from normal navigation) the `Plate type` selector defaults to `Choose plate type...`. See [this Stack Overflow thread](https://stackoverflow.com/questions/43043113/how-to-force-reloading-a-page-when-using-browser-back-button) for some potential solutions; since we have jQuery in this project, there are probably ways of getting around this through the use of minimal code.

(for reference: this error was observed in Chrome 76.0.3809.100 on macOS 10.13.6)

fedarko:
**Note: this issue also happens if you navigate back to the "Plate samples" page. See attached screenshot** (notice how the "Create" button is disabled, even though a valid plate configuration is selected).

<img width="1552" alt="Screen Shot 2019-08-19 at 6 20 38 PM" src="https://user-images.githubusercontent.com/4177727/63310085-2c02f480-c2ae-11e9-98ec-b1e80a68c184.png">

This is a pretty general design issue, so my guess is that it's impacting multiple pages throughout the LabControl GUI—not just the two I've highlighted here.

Issue #553, opened 2019-07-27
Plate-mapper behaving incorrectly.


charles-cowart:
Wet-lab has an issue with blank cells - it appears when the data is all lower-case and is copied and pasted from google sheets, LabControl can reformat the data as intended. If one or more characters are uppercase, LC will highlight the items in red.

Wet-lab still needs to manually input the Qiita Study ID before the Tube ID for the AGP and Palmer Rats studies. This may be related, hence it has been documented here.

Wet-lab has observed an issue while working with Palmer_Rats_11479_35. When migrating away from the Plate-mapper dialog, entries flagged red usually disappear, while normal entries are saved. During recent use, they have experienced the opposite happening at least on one occasion, as seen in the image.

All three items should be investigated as a whole, since the plate-mapper code is highly interactive. Some items may be duplicated in part from a previous issue.

<img width="1191" alt="Screen Shot 2019-07-23 at 7 47 06 AM" src="https://user-images.githubusercontent.com/42684307/61988868-32b68880-afdc-11e9-89b2-13a389d9de4b.png">

charles-cowart:
> Wet-lab has an issue with blank cells - it appears when the data is all lower-case and is copied and pasted from google sheets, LabControl can reformat the data as intended. If one or more characters are uppercase, LC will highlight the items in red.

The above part of this issue was resolved with PR #557

charles-cowart:
Confirmed that with this fix, when control types ('blank', 'contam', etc.) are cut and pasted into a cell or typed, the control type will appear in the pull-down menu, even if the entry is in mixed or upper case. Confirming w/wetlab that when they cut and paste the blank control type, it is just the keyword 'blank' in arbitrary case.

However, I believe the use case where this is an issue is actually when multiple cells have been selected and copied from Excel/Google Sheets and pasted into the plate-mapper. Once we have a solid solution for handling typing/single cell cut and paste versus multi-cell cut and paste, we will need to make sure it can process control types as well.

charles-cowart:
> Wet-lab still needs to manually input the Qiita Study ID before the Tube ID for the AGP and Palmer Rats studies. This may be related, hence it has been documented here.

[...]

PR #569, opened 2019-08-20
Add a checkbox that allows autocomplete-from-paste and multiselect functionality to coexist; resolve duplicate plate issue

### LINKED ISSUE #354: Trigger autocomplete with paste as well as typing

### LINKED ISSUE #567: Multiple plates are being created for a single plating operation
So here's the thing. While working on #354, I noticed that if you put a `console.log()` statement in the `PlateViewer.initialize()` function, then it would be called _twice_ when starting a plating operation: once when you select a plate configuration (e.g. `96-well deep-well plate`), and then once again when you actually press the green "Create" button.

I was under the impression initially that this was being done on purpose for some reason, but after tracking this down (and talking with @charles-cowart) it sounds like this is actually a long-standing bug?

For reference, running `grep -r "new PlateViewer(" labcontrol/gui/static/js/` gives you two hits, both in `plate.js`. Here's the function containing the first one:
 
https://github.com/jdereus/LabControl/blob/ae2b8a5a3315bfed154c617cfda43dbf668462dc/labcontrol/gui/static/js/plate.js#L109-L122

and here's the second one (not copying the entire function due to space, but it's in `createPlate()`):

https://github.com/jdereus/LabControl/blob/ae2b8a5a3315bfed154c617cfda43dbf668462dc/labcontrol/gui/static/js/plate.js#L156

Anyway, as you can see: in the first block of code, `pv` is never used. From what I can tell you can straight-up comment this line out without breaking anything.

There is still room to investigate this further—1) verifying that this is actually a bug, 2) verifying the effects of this if it is a bug (at the very least everything in `PlateViewer.initialize()` is called twice), and 3) verifying that the fix doesn't mess up the front-end—but I think this will be a good place to pick up on Monday.

Tagging @antgonza since Charlie mentioned you'd run into this sort of error with LC before, and tagging @qiyunzhu since this might complicate unit testing a bit until a fix would be merged in.

fedarko:
Closes #354 and closes #567, and represents partial progress on #553. Details:

## Adding a checkbox that allows autocomplete and multiselect functionality to coexist
This PR implements the solution described by @wasade in #464. It's a pretty simple solution, thankfully.

Note that this checkbox defaults to "checked" (i.e. multiselect pasting is enabled, at the expense of the autocomplete popping up on pasting). _However_, if the page loads and the checkbox isn't checked (due to browser back/forward button shenanigans analogous to those causing #562), then my code respects that and holds off on registering the CECM plugin accordingly.

Something worth noting is that testing this on the 384-well plates is an ...interesting experience, because the autocomplete takes a few seconds to activate when pasting something in for a given cell. (But it does show up eventually.) Improving the autocomplete performance seems out of the scope of this PR, however.

## Resolve duplicate plate issue
This removes one line of code in `plate.js`, and thereby closes #567. (I was surprised fixing this was as easy as that, so I'm suspicious that this actually solves the issue that @antgonza and @wasade were apparently experiencing a while back. Not sure how to go about testing that that issue is actually resolved, though.)

## In any case
That's it for now! Thanks in advance for reviewing this PR, and happy to meet to discuss this in person tomorrow if that'd be easier.

### DIFF
--- a/labcontrol/gui/static/js/plateViewer.js
+++ b/labcontrol/gui/static/js/plateViewer.js
@@ -254,7 +254,33 @@ PlateViewer.prototype.initialize = function (rows, cols) {
 
   // don't select the active cell, otherwise cell navigation won't work
   this.grid.setSelectionModel(new Slick.CellSelectionModel({selectActiveCell: false}));
-  this.grid.registerPlugin(new Slick.CellExternalCopyManager(pluginOptions));
+
+  // TODO: This function is called twice every time an event happens, and that
+  // is because this binding is applied twice (because this initialize()
+  // function is called twice). This is likely worth addressing -- there are
+  // multiple ways to circumvent this -- but I'm not sure what would be best.
+  //   Idea 1: add this binding at the bottom of the plate HTML template
+  //   Idea 2: add logic to detect if this binding already exists (buggy?)
+  //   Idea 3: ignore this, and just accept that the callback will be called
+  //          twice (...is this intended?)
+  // NOTE: We use a closure to pass "this" to the callback function
+  var currViewer = this;
+  $('#multiSelectCheckbox').on('change', function() {
+    // In this callback function, "this" is the m.s. checkbox's <input> element
+    if (this.checked) {
+      currViewer.grid.registerPlugin(currViewer.cellExternalCopyManager);
+      // console.log("cecm plugin registered");
+    } else {
+      currViewer.grid.unregisterPlugin(currViewer.cellExternalCopyManager);
+      // console.log("cecm plugin UNregistered");
+    }
+  });
+  // This paradigm inspired by
+  // https://github.com/mleibman/SlickGrid/blob/master/examples/example-spreadsheet.html
+  this.cellExternalCopyManager = new Slick.CellExternalCopyManager(pluginOptions);
+  // TODO: if the user navigates *back* to this page while this plugin is
+  // disabled, I think this will cause problems (analogous to issue #562).
+  this.grid.registerPlugin(this.cellExternalCopyManager);
 
   // When a cell changes, update the server with the new cell information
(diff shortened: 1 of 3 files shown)


labcontrol/gui/static/js/plateViewer.js
```diff
@@ -254,7 +254,37 @@ PlateViewer.prototype.initialize = function (rows, cols) {
 
   // don't select the active cell, otherwise cell navigation won't work
   this.grid.setSelectionModel(new Slick.CellSelectionModel({selectActiveCell: false}));
-  this.grid.registerPlugin(new Slick.CellExternalCopyManager(pluginOptions));
+
+  $('#multiSelectCheckbox').on('change', function() {
+    // In this callback function, "this" is the m.s. checkbox's <input> element
```
charles-cowart: m.s. implies they already understand enough of what they're looking at to know it means multi-select, but it's a minor issue.

fedarko: Ah, that's a fair point! I can update this to be more explicit in the future if you'd like, but I figured it was ok due to the context (since this is already in a callback function for an element called `multiSelectCheckbox`).

PR #571, opened 2019-08-20
chaninging jdereus/labman to biocore/LabControl

### LINKED ISSUE #561: change github url in `setup.py`
https://github.com/jdereus/LabControl/blob/b4ee5c60549c504a0b10c2490669c9c0dfdd970c/setup.py#L39

The `setup.py` currently uses the old url. The old url currently redirects to the new url, so it seems low priority, but I imagine it would be ideal to update this string to 'https://github.com/jdereus/labcontrol'

gwarmstrong:
Fixes #561 . Also fixes the other references to `jdereus/labman` (Thanks @fedarko !)

One quick question:
Should this line be changed to have `biocore` instead of `jdereus` ?
https://github.com/biocore/LabControl/blob/d4de4188123e472a5a54d0aac6e00d179d733fd1/LICENSE#L3


charles-cowart:
Approved. Will merge if/when CI tests are complete.

Issue #171, opened 2018-01-30
plateViewer.js is untested


ElDeveloper:
It would be good to add tests for at least the PlateViewer class, but really anything that may make sense would be good to have properly tested.

charles-cowart:
Since Javascript implements a lot of use-facing functionality, it would be really good to have at least some basic unit-testing for plate mapping at the least. We may find the source of some of our intermittent errors here.

wasade:
This is an ongoing issue partially addressed with the "Establish a Javascript unit testing framework" milestone. Removing from that milestone as fully resolving this specific issue was not a requirement for completion of the milestone.

charles-cowart:
@fedarko @qiyunzhu Now that we have a framework in place to develop additional tests, we can look at adding the most obvious ones for the next release. Following that, we can close this issue and issue individual issues for subsequent tests.

charles-cowart:
@qiyunzhu As mentioned, here is the official issue for the adding unit tests and refactoring javascript code as needed to support those tests. As this issue is relatively general, if you want, we can sit down together and look through the javascript and break this issue down by file, class, etc.

PR #572, opened 2019-08-20
added unit tests for several simple labcontrol functions

qiyunzhu:
@charles-cowart #171 Just a test of QUnit. Nothing heavy.


charles-cowart:
Although this build failed, it looked like it failed an on an earlier version of the shotgun prep file generation test; which was fixed. @qiyunzhu, I'll work with you tomorrow to pull down the latest source and see if we can't integrate your tests into it.

qiyunzhu:
@charles-cowart Thanks! We can discuss tomorrow on this topic!

PR #573, opened 2019-08-20
Address two #562 bugs; integrate "prettier" autoformatter for JS and CSS code

fedarko:
We can swap out prettier / change configs / etc. really easily, if desired (all that is needed is updating the Makefile, which I added as part of this PR).

This PR *looks* hairy because it includes a lot of changes. However, the bulk of these are just automatic changes from running `make festyle` (and this is prefaced by doing a dry run with the `--debug-check` option, to check that stuff won't get broken by running `prettier --write`).

**For reviewers**, I'd suggest mainly just looking at the non-auto-formatter changes (which should be localized to the Travis YAML, the newly added Makefile, and plate.html + plate_list.html). This PR doesn't run prettier on the HTML yet (for reasons explained in the Makefile), so all of these changes aren't the result of prettier autoformatting everything.

**For running this autoformatter locally**, you can just run `npm install -g prettier` to install prettier on your system globally (this has been integrated into the Travis build process), then run `make festyle` from within the root directory of your LabControl repository to do autoformatting. You can also run `make festylecheck` to see if files would require autoformatting changes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,8 @@ before_install:
   # headless-chrome and qunit.
   # node-qunit-puppeteer installs its own chromium as a dependency.
   # hence, a separate chrome installation is not required.
-  - npm install -g node-qunit-puppeteer
+  # ...also, install "prettier" for JS/CSS/HTML auto formatting.
+  - npm install -g node-qunit-puppeteer prettier
 install:
   # Hard limit for transaction log, to prevent VHD from filling during tests
   - psql -c "ALTER SYSTEM SET max_wal_size=2"
@@ -95,6 +96,7 @@ script:
   - python labcontrol/gui/js_tests/js_tester.py all-tests
   - nosetests --exclude-dir=labcontrol/gui/js_tests --with-doctest --with-coverage -v --cover-package=labcontrol
   - flake8 labman setup.py scripts/*
+  - make festylecheck
   - labcontrol/db/tests/tester.py integration_tests
   - labcontrol/db/tests/tester.py stress_tests --num_plates_amplicon 2 --num_plates_shotgun 2
   - labcontrol start-webserver &
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ before_install:
   # headless-chrome and qunit.
   # node-qunit-puppeteer installs its own chromium as a dependency.
   # hence, a separate chrome installation is not required.
-  # ...also, install "prettier" for JS/CSS/HTML auto formatting.
+  # ...also, install "prettier" for front-end code auto formatting.
   - npm install -g node-qunit-puppeteer prettier
 install:
   # Hard limit for transaction log, to prevent VHD from filling during tests
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,27 @@
+# NOTE 1: these targets require the "prettier" Node.js package is globally
+# installed
+#
+# NOTE 2: this code was derived from Qurro's Makefile
+# (https://github.com/biocore/qurro/blob/master/Makefile)
+#
+# NOTE 3: HTML code is currently not included in these operations, since it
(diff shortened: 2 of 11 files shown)


.travis.yml
```diff
@@ -31,7 +31,8 @@ before_install:
   # headless-chrome and qunit.
   # node-qunit-puppeteer installs its own chromium as a dependency.
   # hence, a separate chrome installation is not required.
-  - npm install -g node-qunit-puppeteer
+  # ...also, install "prettier" for front-end code auto formatting.
```
charles-cowart: '...' isn't needed, but that's just my opinion.

.travis.yml
```diff
@@ -95,6 +96,7 @@ script:
   - python labcontrol/gui/js_tests/js_tester.py all-tests
   - nosetests --exclude-dir=labcontrol/gui/js_tests --with-doctest --with-coverage -v --cover-package=labcontrol
   - flake8 labman setup.py scripts/*
+  - make festylecheck
```
charles-cowart: if anything, add another comment here saying this like performs the style check; since it's not the same name as prettier, it's not immediately obvious to anyone who isn't seeing the two lines highlighted.

Makefile
```diff
@@ -0,0 +1,27 @@
...
+# but some of that seems to be due to the Tornado template syntax messing
+# things up.)
+.PHONY: festyle festylecheck
+
+JS_CSS_CODE_LOCS = labcontrol/gui/static/js/*.js labcontrol/gui/static/css/labcontrol.css labcontrol/gui/js_tests/*.js
+HTML_CODE_LOCS = labcontrol/gui/templates/*.html labcontrol/gui/js_tests/*.html
+
+# This target auto-formats all of the JS and CSS code to be compliant with
+# prettier.
+festyle:
```
charles-cowart: Might want to say what festyle stands for, if it's a label.

labcontrol/gui/js_tests/addPlateModal_tests.js
```diff
@@ -5,150 +5,162 @@
...
 
 // Tests of functions that depend on the plate modal div having already
 // been inserted
-QUnit.module("modal-div-dependent tests", function (hooks) {
-    hooks.beforeEach(function (assert) {
-        insertPlateModalDiv("qunit-fixture");
+QUnit.module("modal-div-dependent tests", function(hooks) {
+  hooks.beforeEach(function(assert) {
+    insertPlateModalDiv("qunit-fixture");
+  });
```
charles-cowart: @fedarko Just to confirm, there is supposed to be a closing brace here, before ');'? I'm asking because this has the effect of taking everything that was in beforeEach() below, and moving it out.

labcontrol/gui/static/js/addPlateModal.js
```diff
@@ -57,86 +64,95 @@ function insertPlateModalDiv(elementIdToAppendTo = undefined,
...
+        render: function(data, type, row, meta) {
+          var plateId = data[0];
+          return (
+            "<button id='" +
+            specificAddPlateBtnBaseId +
+            plateId +
+            "' class='btn btn-success btn-circle-small'>" +
+            "<span class='glyphicon glyphicon-plus'></span>" +
+            "</button>"
+          );
```
charles-cowart: Note for other readers - the prettifier has encapsulated return's params in an (), and this affected formatting.

charles-cowart:
@gwarmstrong if it's reasonable, you can approve it. However, I'll manage the merging of the code, and make sure it's done after it's passed CI integration.

labcontrol/gui/js_tests/addPlateModal_tests.js
fedarko: Note for future reference: @charles-cowart and I looked at the history of this file and verified that these changes were ok.

Issue #576, opened 2019-08-21
Turn `Pools` -> `Pool library plates` into two separate menus, one for shotgun and one for amplicon


gwarmstrong:
@charles-cowart told me about this issue. The main idea is that the `Pool library plates` menu currently makes users choose `Shotgun library paltes` or `Amplicon library plates`, and then they can add plates. For this issue, we would like to split these into two separate dropdowns from `Pools`.

It also looks like there are some inconsistencies between drop-downs for `Plates`, `Pools`, and `Sequencing Runs`.
Namely, `List` being in its own container for the first two, but combined in with the other drop-down options in `Sequencing Runs`. Also, it might be nice to break `Pools` and `Sequencing Runs` up into `Amplicon workflow` and `Shotgun workflow`, as in the `Plates` menu.

gwarmstrong:
Working on this on https://github.com/biocore/LabControl/tree/issue576-separate-pool-library-plates

PR #579, opened 2019-08-23
Issue576 separate pool library plates

gwarmstrong:
Addressing #576 

Discussing with Charlie, this issue extends beyond the GUI, to totally separating the templates and handlers for the shotgun pooling and amplicon pooling.

At this point I have changed the gui to have separate interfaces, and use separate templates, but they still rely on the same handler.

Some demos of the new interface:
Screen Shot 2019-08-23 at 11 15 22 AM
Screen Shot 2019-08-23 at 11 15 41 AM

Note: there may be some misc. commented out code in the diff that I need to remove


charles-cowart:
@gwarmstrong This looks good - it looks like the decision making for 16S vs Shotgun pooling has been centered into pooling.py. Now the aim is to reduce that complexity in pooling.py. @fedarko what do you think? Unless Marcus sees issues to do otherwise, I may just approve this as is, and immediately issue another PR with a refactored pooling.py.

labcontrol/gui/handlers/process_handlers/test/test_pooling_process.py
```diff
@@ -65,25 +65,74 @@ def test_post_pool_pool_process_handler(self):
...
+
+        # this process id does not exist
+        response = self.get('/process/poollibraries/shotgun_plate/'
+                            '?process_id=10000')
         self.assertEqual(response.code, 404)
 
     def test_post_library_pool_process_handler(self):
 
+        # TODO may want to include test to make sure it is not able to post
+        #  plates with the wrong type
```
fedarko: this is the most minor issue possible but there's an extra space before "plates"

labcontrol/gui/templates/library_pooling_16S.html
```diff
@@ -71,25 +68,17 @@
...
-      return 15;
-    }
-    else if (plateType == 'shotgun library prep') {
-      return 1000;
-    }
-
-    // should never get here but if it does, then the range is permissive
-    return 10000;
+  function clippingForLibraryPlateType() {
+    // default value ofr 16S
```
fedarko: Should be `for` instead of `ofr`, as far as I can tell

charles-cowart:
In general, it is good to have documentation be tidy. Good catch, Marcus! I
can take care of it on Sunday - I'm going to make a pass through the
pooling_process.py file, to split the pathways for 16S and shotgun, and
make a new one for MiniPCR.

labcontrol/gui/handlers/process_handlers/test/test_pooling_process.py
gwarmstrong: I usually put a second space in continuation lines of comments. I like this convention, but feel free to change.

labcontrol/gui/templates/library_pooling_16S.html
gwarmstrong: Good call.

PR #582, opened 2019-08-27
Pool sep

charles-cowart:
First pass at encapsulating PoolHandler child-classes. Note we can do more to derive the common elements in _compute_pools().


charles-cowart:
Yeah, caught that after reviewing the last test output. I need to check if
matrix was just a holdover from incorporating the latest qiita version or
not.

PR #587, opened 2019-09-05
change labman to labcontrol in .travis.yml

### LINKED ISSUE #586: flake8 does not check labcontrol directory
https://github.com/biocore/LabControl/blob/1427ad163682ff11b1086e0e3b1010ba1a2166ee/.travis.yml#L100

This line has `labman` instead of `labcontrol` and flake8 does not throw non-zero exit codes if it is passed directories/files that do not exists. So this should be changed to `labcontrol` so the python code is actually linted.

gwarmstrong:
Fixes #586 . 

Changes:
change `labman` to `labcontrol` in flake8 checks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -97,7 +97,7 @@ install:
 script:
   - python labcontrol/gui/js_tests/js_tester.py all-tests
   - nosetests --exclude-dir=labcontrol/gui/js_tests --with-doctest --with-coverage -v --cover-package=labcontrol
-  - flake8 labman setup.py scripts/*
+  - flake8 labcontrol setup.py scripts/*
   # Run prettier to verify that the JS/CSS ("front end") code doesn't need any
   # stylistic adjustments.
   - make festylecheck
--- a/.travis.yml
+++ b/.travis.yml
@@ -97,7 +97,7 @@ install:
 script:
   - python labcontrol/gui/js_tests/js_tester.py all-tests
   - nosetests --exclude-dir=labcontrol/gui/js_tests --with-doctest --with-coverage -v --cover-package=labcontrol
-  - flake8 labcontrol setup.py scripts/*
+  - flake8 labcontrol setup.py scripts/* --max-line-length 120
   # Run prettier to verify that the JS/CSS ("front end") code doesn't need any
   # stylistic adjustments.
   - make festylecheck
--- a/.travis.yml
+++ b/.travis.yml
@@ -97,7 +97,7 @@ install:
 script:
   - python labcontrol/gui/js_tests/js_tester.py all-tests
   - nosetests --exclude-dir=labcontrol/gui/js_tests --with-doctest --with-coverage -v --cover-package=labcontrol
-  - flake8 labcontrol setup.py scripts/* --max-line-length 120
+  - flake8 labcontrol setup.py scripts/*
   # Run prettier to verify that the JS/CSS ("front end") code doesn't need any
   # stylistic adjustments.
   - make festylecheck
--- a/labcontrol/db/composition.py
+++ b/labcontrol/db/composition.py
@@ -7,7 +7,6 @@
 # ----------------------------------------------------------------------------
 
 import re
-from enum import Enum
 from . import base
 from . import sql_connection
 from . import process
@@ -1114,7 +1113,6 @@ class PoolComposition(Composition):
             assay_type = TRN.execute_fetchlast()
 
             return assay_type
-
 
     @staticmethod
     def get_components_type(components):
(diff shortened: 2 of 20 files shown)


gwarmstrong:
Heads up, the first build will probably fail since there are some style violations sprinkled throughout the python code ~100 or so in total, when I checked locally. Will update, but wanted to get this in first.

qiyunzhu (approved):
Looks good!

fedarko (approved):
Looks good to me.

AmandaBirmingham:
Hi, @gwarmstrong re https://github.com/biocore/LabControl/pull/587#issuecomment-528491459, are you saying you'd like to merge this with the flake8 failures and then fix them in a subsequent PR, or have I misunderstood?

gwarmstrong:
@AmandaBirmingham thanks for clarifying. Let's fix it here, since it will just cause problems for people later if we don't. Just haven't gotten around to fixing the ~100 lines or so yet.

gwarmstrong:
@AmandaBirmingham @fedarko Okay this should be good to review now. Note that I increased the `max-line-length` for `flake8` since most of the lines violating the `line-length` of 79 were SQL queries in python, and I felt like it increased the readability by allowing longer lines, as compared to forcing all queries to abide by the typical python line-length limits.

fedarko (approved):

.travis.yml
```diff
@@ -97,7 +97,7 @@ install:
 script:
   - python labcontrol/gui/js_tests/js_tester.py all-tests
   - nosetests --exclude-dir=labcontrol/gui/js_tests --with-doctest --with-coverage -v --cover-package=labcontrol
-  - flake8 labman setup.py scripts/*
+  - flake8 labcontrol setup.py scripts/* --max-line-length 120
```
AmandaBirmingham: Sorry--The 80-character line length limit (the `flake8` default) is standard in Knight Lab software (e.g., Qiita, deblur, EMPeror, biom-format, etc), so we really can't drop it unilaterally. Pls regress that change.

AmandaBirmingham (changes requested):
We are going to need to stick to the 80-char line limit (see longer comment in review); sorry for the extra pain :-/

.travis.yml
gwarmstrong: My impression was this is the case because 79-characters is PEP8. The main reason for abiding PEP8 being readability and stylistic consistency. In cases such as this, I am tempted to invoke [this](https://www.python.org/dev/peps/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds) part of PEP8.  In my opinion, aggressively line-splitting in the case of some of these SQL queries can reduce readability.

That being said, if you insist, I am happy to oblige.

AmandaBirmingham: I tend to agree with you about line-splitting the SQL statements reducing readability!  However, my understanding (correct me if I'm wrong!) is that setting the max line length of flake8 to 120 will mean it will not alert on ANY line in the code that is <120 lines long, including the many that are pure python and should stay at the 80-character limit for consistency with other projects.  You could, though, mark of the SQL statements that you want to have longer lines with `# noqa: E501` (see http://flake8.pycqa.org/en/3.1.1/user/ignoring-errors.html#in-line-ignoring-errors ), which would prevent them from causing flake8 warnings while still letting the rest of the codebase follow the shared convention.

gwarmstrong: > setting the max line length of flake8 to 120 will mean it will not alert on ANY line in the code that is <120 lines long, including the many that are pure python and should stay at the 80-character limit for consistency with other projects.
I believe this is correct as well.
>  You could, though, mark off the SQL statements that you want to have longer lines with # noqa: E501 (see http://flake8.pycqa.org/en/3.1.1/user/ignoring-errors.html#in-line-ignoring-errors ), which would prevent them from causing flake8 length warnings while still letting the rest of the codebase follow the shared convention.
Oh that's cool, I haven't seen that before, thanks! 
That being said, I think in this instance, the `# noqa: E501` solution is less than 'developer-friendly' solution if these lines ever need to be updated, and for future instances of _a little too long for your average python line_ SQL queries. Unless your line lengths are setting off red-flags for the reviewer, I feel that hard limit at 120, self-enforced soft limit at 79 is a reasonably provident solution. Feel free to disagree, we can chat at the meeting later :)

gwarmstrong: I talked with @AmandaBirmingham and we were not sure if we have the authority to increase the python line-length limit for this project. @antgonza or @ElDeveloper can either of you comment on whether this seems reasonable?

gwarmstrong:
@AmandaBirmingham let's go ahead and merge this sans feedback on line length if everything else checks out. I have adjusted line-length accordingly.

AmandaBirmingham (approved):

